PR #4, opened 2022-01-13
Fs 391 starter tests

SentientHousePlant:
I branched this off FS-389 - so expect some pipeline stuff in this which will be overwritten by FS-389 later on.

This was to confirm that my tests worked with GitHub Actions.

I have created some very basic tests which essentially test that flask spins up and serves the "Hello World" webpage.

I have a suspicion that more tests may be needed - but I will leave this up for team discussion during this pull request.


Adam-W1 (changes requested):
Can we include how to initiate the test suite in the README (i.e run 'python -m pytest' at the top level directory)

tests/route_testing_conf.py
```diff
@@ -0,0 +1,7 @@
+"""
+A very basic python configuration file.
+Our single source of truth for which
+routes need to be tested and their expected
+content.
+"""
+routes_and_test_content = {"/": b"Hello World", "/404": b"Page not found"}
```
harryyo, line 7: Very clean implementation - for future use (and discussion) perhaps this should be updated to enable specific tags to be checked for content potentially ?

stevewalton28 (approved):

Adam-W1 (approved):

PR #5, opened 2022-01-17
FS-432 WCAG checks in pipeline

harryyo:
Passing accessability tests using Axe, Selenium, Pytest, pytest-flask, pytest-selenium and axe-selenium-python. Axe reporting one minor aria role violation from a govUK-frontend component. create_app initialisation converted to callable function with routes loaded via Blueprint to enable calling from test fixtures. New Config class in config.py for easier config attribute management, and error_routes moved to separate file to allow app error_handler decorators without circular import when included within Blueprint routes.


app/config.py
```diff
@@ -0,0 +1,17 @@
+"""Flask configuration."""
+from os import environ
+from os import path
+
+from dotenv import load_dotenv
+
+basedir = path.abspath(path.dirname(__file__))
```
SentientHousePlant, line 7: Would this piece of code lead to one storing the .env file in the /app directory? 

I think that typically configs are stored in the root, so that configuration and source code are separated. Although that is a bigger concern if one is building the project into a wheel or some other package file - since you don't want to include configuration in there.

What are the benefits of using a class to store configuration over a yaml/toml file?

app/create_app.py
```diff
@@ -1,44 +1,85 @@
...
+        # app.config.from_pyfile("config.py", silent=True)
+        app.config.from_object(Config())
+    else:
+        # load the test config if passed in
+        app.config.update(test_config)
+
+    # ensure the instance folder exists
+    try:
+        os.makedirs(app.instance_path)
+    except OSError:
```
SentientHousePlant, line 36: There is numerous reasons for python to throw a OSError, please see the functions definition [here](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/os.py) . 

You can instead pass in exists_ok=True, this will remove the need for a try - except.

app/config.py
harryyo: Yes agreed and good spot, either this file (and definitely the .env file if used) should be in the root. The benefits of a class are that you can have a base config and then inherit the base config in different classes which automatically get loaded in different environments (with perhaps just one prod/dev environment set to toggle between them. However having lots of different environment variables set in the application at the moment is overkill. So perhaps easiest for now is to remove this (also as it is unused) and just take the environment variables from the global environment or defaults if set

app/create_app.py
harryyo: Good spot and nice suggestion - done!

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+    from app.routes import bp as default_routes
+
+    app.register_blueprint(default_routes)
+
+    return app
+
+
+app = create_app()
 
-from app import routes  # noqa
```
Adam-W1, line 44: As discussed elsewhere - routes need to be created via blueprint within the create_app function for invocation as part of test fixtures (this would not happen if they were imported outside of create_app as error_routes currently are). We will therefore need to instantiate error routes within a blueprint to test them at some point.
*Flask requires a special work around for routes involving 404 error pages as the 404 occurs at the routing level before the blueprint can be determined.

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+
+    @app.context_processor
+    def inject_global_constants():
+        return dict(
+            stage="alpha",
+            region="NA",
+            service_title="DLUHC Funding Service Design Iteration 1",
+            service_meta_description="DLUHC Funding Service Design Iteration 1",  # noqa
+            service_meta_keywords="DLUHC Funding Service Design Iteration 1",
+            service_meta_author="Evoco Digital Services",
```
stevewalton28, line 69: and DLUCH

stevewalton28 (changes requested):

app/create_app.py
harryyo: New custom error_handlers defined for 404 and 500 within create_app initialiser to enable unified handling and testing of error routes alongside other application defined routes

app/create_app.py
stevewalton28: Yes

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+        )
+
+    from app.routes import bp as default_routes
+    from app.routes import not_found, internal_server_error
+
+    app.register_error_handler(404, not_found)
+    app.register_error_handler(500, internal_server_error)
```
Adam-W1, line 76: i think internal_server_error needs renaming to internall_server (or vice versa) and the import breaks locally

app/create_app.py
harryyo: Done

app/create_app.py
harryyo: Apols, thanks for this Adam. Renamed all to internal_server_error

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
-        service_meta_description="DLUHC Funding Service Design Iteration 1",
-        service_meta_keywords="DLUHC Funding Service Design Iteration 1",
-        service_meta_author="Evoco Digital Services",
+from jinja2 import ChoiceLoader
+from jinja2 import PackageLoader
+from jinja2 import PrefixLoader
+
+
+def create_app(test_config=None):
+    # create and configure the app
```
endamccormack, line 13: Have we agreed this level of code commenting? Its a bit much. 

If its not been agreed then I'd say make a new branches with the comments for later discussion and then remove them all from this pull request. 

Setup a session for us to discuss at some stage this week

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+    app = Flask(
+        __name__, instance_relative_config=True, static_url_path="/assets"
+    )  # noqa
+
+    app.config.from_mapping(
```
endamccormack, line 18: I dont think we should be doing this. You dont tend to change config on the fly like this. You can use different config sources for dev, test, prod etc

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+        # a default secret that should be overridden by instance config
+        SECRET_KEY="dev",
+        # store the database in the instance folder
+        DATABASE=os.path.join(app.instance_path, "flaskr.sqlite"),
+    )
+
+    if test_config is None:
```
endamccormack, line 25: does this mean config for test or do we mean something else? Generally, It's generally frowned upon to add additional code that adapts for your testing as it can change the baseline of how the application acts vs how it would in the real world

app/create_app.py
```diff
@@ -1,44 +1,82 @@
...
+        # load the instance config, if it exists, when not testing
+        # app.config.from_pyfile("config.py", silent=True)
+        app.config.from_object(Config())
+    else:
+        # load the test config if passed in
+        app.config.update(test_config)
+
+    # ensure the instance folder exists
+    os.makedirs(app.instance_path, exist_ok=True)
```
endamccormack, line 34: is there any point in this line, if it is not there it will error either way right?

app/create_app.py
```diff
@@ -1,44 +1,82 @@
+import os
```
endamccormack, line 1: file in general is a bit messy, think about how you might break out some of this or group it differently to make it tidier

tests/test_accessibility.py
```diff
@@ -0,0 +1,66 @@
...
+from axe_selenium_python import Axe
+from flask import url_for
+
+multiprocessing.set_start_method("fork")  # Required on macOSX
+
+
+@pytest.mark.app(debug=False)
+def test_app(app):
+    assert not app.debug, "Ensure the app not in debug mode"
+
```
endamccormack, line 18: make spacing between methods common and classes common

endamccormack (changes requested):

harryyo:
Removing superfluous config mapping from create_app

app/create_app.py
harryyo: Removed for now as obvious

app/create_app.py
harryyo: Removed on the fly config mapping in pref for global environment config management

app/create_app.py
harryyo: Was included as a location for on the fly fixtures so removed for now as redundant

app/create_app.py
harryyo: Removed unnecessary on the fly config management

app/create_app.py
harryyo: Removed in favour of global environment config management

tests/test_accessibility.py
harryyo: Spacing made consistent

vialde (approved):
These changes look sufficient

tests/test_accessibility.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        res = urlopen(url_for("routes.index", _external=True))
+        assert b"Hello World" in res.read()
+        assert res.code == 200
+
+
+@pytest.mark.usefixtures("selenium_chrome_driver")
+@pytest.mark.usefixtures("live_server")
+class TestURLsWithChrome:
+    def test_driver_can_open_url(self):
```
stevewalton28, line 42: Why do we need to test that we can open google ?

stevewalton28 (changes requested):

tests/test_accessibility.py
harryyo: This was to test selenium was working, but not needed now - removed

stevewalton28 (approved):

endamccormack (approved):

vialde:
I've added the report artifact upload to the pipeline so you get a .zip of all Harry's lovely tables

tests/test_accessibility.py
```diff
@@ -0,0 +1,118 @@
...
+            or results["violations"][0]["impact"] == "minor"
+        )
+
+    def test_unknown_page_returns_accessible_404(self):
+        """
+        GIVEN Our Flask Application is running
+        WHEN the '/page-that-does-not-exist' page is requested (GET)
+        THEN check that a 404 page that is returned conforms to WCAG standards
+        """
+        # route = url_for("routes.index", _external=True) + "rubbish"
```
endamccormack, line 105: this needs to be addressed

Adam-W1, line 105: wondering if this comment is needed?

endamccormack (changes requested):

tests/test_accessibility.py
harryyo: Apols, yes  removed

harryyo: Apols, yes, removed

endamccormack (approved):

PR #7, opened 2022-01-17
FS-389 Initial Pipeline

vialde:
Need to get these pipeline assets over to 'Main' so the devs can start using them.


SentientHousePlant (commented):
Looking forward to seeing the accessibility tests pass with this, lovely stuff! I too wouldn't be able to push with failing tests.

SentientHousePlant:
Only comment would be that in the future it will be nice to have a cache (as discussed) which holds the build files so we don't need to run build.py twice.

A question for the team : Would there ever be an instance where one would want to test without deploying? Not sure if GitHub Actions supports conditional jobs, but just a (potentially useless) thought.

README.md
```diff
@@ -2,6 +2,8 @@
 
 [![made-with-python](https://img.shields.io/badge/Made%20with-Python-1f425f.svg)](https://www.python.org/)
 
+![Hello World Deploy](https://github.com/communitiesuk/funding-service-design-frontend/actions/workflows/govcloud.yml/badge.svg)
```
stevewalton28, line 5: This is going to stop being hello-world next sprint - probably call it the funding-service-design-frontend ?

stevewalton28 (changes requested):

README.md
vialde: sure

stevewalton28 (approved):

endamccormack (approved):

PR #20, opened 2022-02-04
XGov Forms Runner building, and tests running on instance and passing

harryyo:


.github/workflows/govcloud.yml
```diff
@@ -43,109 +126,109 @@ jobs:
           name: accessibility-test-report
           path: /home/runner/work/funding-service-design-frontend/funding-service-design-frontend/axe_reports/*.html
           retention-days: 5
+# #######
```
stevewalton28, line 129: Is this commented out?  If it is, why?

.github/workflows/xforms_builder.yml
```diff
@@ -23,8 +23,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      -
-        name: Set up Docker Buildx
+      - name: Set up Docker Buildx
```
stevewalton28, line 26: "Buildx" ? typo?

tests/test_application_form.py
```diff
@@ -0,0 +1,262 @@
+"""
```
stevewalton28, line 1: This is a massive comment at the beginning of the file - is it necessary?  What's the logic of putting this here over a wiki or something else?

stevewalton28 (changes requested):

.github/workflows/xforms_builder.yml
vialde: nope, it's actually called buildx

.github/workflows/govcloud.yml
vialde: Yeah, it's commented out and due to be removed....I'm going to resolve this in another branch.  little bit of cross coding happening as we bed in the workflows

tests/test_application_form.py
harryyo: Sure I agree - the json schema is tightly coupled with the tests so thought worth having a reference somewhere near the code, in case the schema changes and the tests fail as a result - I can slim down definitely as a first step... But would welcome suggestions on better places for it

harryyo: Schema removed and comment refined

.github/workflows/govcloud.yml
harryyo: Commented code removed

stevewalton28 (approved):

SentientHousePlant (approved):
Looks good, glad to see a proper config file. 

Will be good to swap out the tests/route_testing_conf.py for a proper dynamic route finder.

manifest.yml
```diff
@@ -11,3 +11,12 @@ applications:
   buildpacks:
   - https://github.com/cloudfoundry/python-buildpack.git
   command: flask run --host 0.0.0.0 --port 8080
+
+- name: funding-service-design-form-runner
+  docker:
+    image: ghcr.io/communitiesuk/digital-form-builder/runner:latest
+  memory: 528M
+  sandbox: true
+  # route: funding-service-design-form-runner.london.cloudapps.digital
```
Adam-W1, line 20: can this be removed?

harryyo: Yes, can be removed

PR #22, opened 2022-02-08
Security job inclusion

vialde:


Adam-W1 (approved):

.github/workflows/govcloud.yml
```diff
@@ -126,3 +126,9 @@ jobs:
           name: accessibility-test-report
           path: /home/runner/work/funding-service-design-frontend/funding-service-design-frontend/axe_reports/*.html
           retention-days: 5
+      - name: ZAP Scan
+        uses: zaproxy/action-full-scan@v0.3.0
+        with:
+          target: ${{ secrets.FORMS_SERVICE_HOST }}
+          allow_issue_writing: False
+
```
harryyo, line 134: It seems the ZAP action by default doesn't exit with an error (even if high security alerts) unless you set it to with this flag - so this should be added at the bottom `fail_action: true`

SentientHousePlant (approved):

PR #32, opened 2022-02-22
Fs 537 submit application feature with summary page & submit button (Fails Accessibility Test on Summary page)

ramsharma-prog:


tests/test_application_form.py
```diff
@@ -106,8 +106,8 @@ def test_form_pages_accessible(self):
                     },
                     route_rel,
                 )
-                assert len(results["violations"]) <= 2
-                assert len(results["violations"]) == 0 or all(
+                assert len(results["violations"]) <= 4
+                assert len(results["violations"]) <= 4 or all(
                     [
```
harryyo, line 111: This will allow four serious/critical accessibility violations. The causes of the serious/critical violations should really be fixed where they're occurring rather than allowing them through here.

form_jsons/public/funding-application.json
```diff
@@ -241,6 +242,9 @@
         }
       ],
       "next": [
+        {
+          "path": "/dummy-page"
+        }
       ]
```
harryyo, line 248: Is there no way to have multiple pages redirecting to the summary (rather than "dummy page" before the summary)? Have we established what the constraints are for page sequences? We should let the design team know if we need to have certain pages in sequence (or otherwise perhaps raise a ticket to configure the runner to not require a "dummy page")

harryyo (changes requested):

tests/test_application_form.py
```diff
@@ -106,8 +106,8 @@ def test_form_pages_accessible(self):
                     },
                     route_rel,
                 )
-                assert len(results["violations"]) <= 2
-                assert len(results["violations"]) == 0 or all(
+                assert len(results["violations"]) <= 4
```
SentientHousePlant, line 109: Might be worth having a configuration file where stuff like this is specified - in the future we then have a single source of truth for where we are allowing problems through our pipeline.

form_jsons/public/funding-application.json
SentientHousePlant: + 1

tests/test_application_form.py
SentientHousePlant: This pull request is uncontroversial but nevertheless it raises some questions about how we've handled a11y violations - I think a test config should be created and used to track anywhere that violations are allowed because of third party libraries. We can also create an issue in the x-gov-form-builder which states where is a serious a11y violation in the summary page.

ramsharma-prog: The accessibility issue is coming from the summary page. I  had communicated with Jen for further advice & she has advised raising a ticket so she can work on it.

form_jsons/public/funding-application.json
ramsharma-prog: If multiple pages are directly linked to the summary page, it doesn't allow to submit the summary page. it’s protection against a user skipping to the summary page and just clicking submit or we could have four separate summary pages linked to each of the pages from the country list instead of a dummy page which works the same.

tests/test_application_form.py
ramsharma-prog: I have  logged the issue on Github: https://github.com/XGovFormBuilder/digital-form-builder/issues/765 & 

raised a ticket on Jira: https://digital.dclg.gov.uk/jira/browse/FS-629

SentientHousePlant: Great Work Ram! Very detailed Jira ticket, I will add to the X-Gov-Form ticket on where to fix the issue!

ramsharma-prog: Thank you, Jack

harryyo: Great tickets/issues Ram. This line still allows four serious accessibility violations **per page** though. Is this what we want to do? If needed for now to finish this sprint, perhaps we should add a ticket in Jira to note the fact that the accessibility tests on this main repo currently allow (up to 4) serious accessibility violations per page? (So we don't forget to bring the threshold down later?)

ramsharma-prog: Thanks, Harry, with this we can see the summary page & the submit button & if I remove the summary page, it means the submit button would be removed however the application would still be submitted but from the user's point of view, I think the submit & summary page is important? I have already raised the ticket on Jira & logged the issue on X-Gov GitHub & communicated with Jen. I will keep chasing this issue with Jen.

form_jsons/public/funding-application.json
ramsharma-prog: So somehow we don't need a dummy page anymore. I have updated the JSON :)

tests/test_application_form.py
ramsharma-prog: Thanks Harry, I have raised the ticket on Jira & logged an issue on X-Gov Github, communicated with Jen & I will keep chasing with Jen

ramsharma-prog: I have raised an issue on Github with ref: https://github.com/communitiesuk/funding-service-design-frontend/issues/34

Also, I have raised a ticket on Jira: https://digital.dclg.gov.uk/jira/secure/RapidBoard.jspa?rapidView=110&view=planning&selectedIssue=FS-636&issueLimit=100

harryyo: Nice work Ram

harryyo (approved):
Nice Ram, well resolved, well documented.

stevewalton28 (approved):

PR #36, opened 2022-04-22
Fs 764 save and return

Adam-W1:


app/default/routes.py
```diff
@@ -24,6 +30,39 @@ def not_eligible():
...
+@default_bp.route("/continue_application", methods=["GET", "POST"])
+def continue_application():
+    """
+    Page to return to an active application
+    Provides a way of returning to an compelte/incomplete
+        application to complete or make changes
+    The user provides an application id and is redirected to
+        the application summary page within the runner
+    """
+    form = FlaskForm()
```
harryyo, line 42: Maybe this should be broken out into a form class somewhere, so we can fill it out with validation for the application_id and any other required fields in future. The following line can then use if form.validate_on_submit() to check validation before using in get_application_data

app/default/data.py
```diff
@@ -0,0 +1,77 @@
...
+    applications_response = get_data(applications_endpoint)
+    return applications_response
+
+
+def extract_data_from_application(application_data, data_subset):
+    data_subset = application_data[data_subset]
+    return data_subset
+
+
+def format_application_data(application_data, form_id):
```
harryyo, line 50: Could this have a more descriptive name like format_application_rehydrate_data? or something like that so obvious what it's used for?

app/default/helpers.py
```diff
@@ -0,0 +1,16 @@
+import requests
+from app.config import FORMS_SERVICE_PUBLIC_HOST
+
+
+def get_return_to_runner_token(form_name: str, formatted_application_payload):
```
harryyo, line 5: maybe this could be get_return_to application_token? (runner could be an ambiguous term)

harryyo (commented):
Great Adam, just a couple of thoughts re naming of functions and perhaps a class for the form?, well done for getting it working!

.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
...
         name: Reorder Python imports (src, tests)
         args: ["--application-directories", "src"]
--   repo: https://github.com/Lucas-C/pre-commit-hooks-bandit
-    rev: v1.0.5
-    hooks:
-    -   id: python-bandit-vulnerability-check
+# -   repo: https://github.com/Lucas-C/pre-commit-hooks-bandit
+#     rev: v1.0.5
+#     hooks:
+#     -   id: python-bandit-vulnerability-check
```
SentientHousePlant, line 29: 📌 We should probably outsource this file to the templates :)

app/default/data.py
```diff
@@ -0,0 +1,77 @@
...
+APPLICATION_ENDPOINT = "/applications/{application_id}"
+
+
+def get_data(endpoint: str):
+    if endpoint[:8] == "https://":
+        response = requests.get(endpoint)
+        if response.status_code == 200:
+            data = response.json()
+        else:
+            return None
```
SentientHousePlant, line 18: 🤔 🔧 Maybe this should be its own encapsulated function - I would expect to have get_local_data, get_remote_data, and finally a get_data_handler function.

app/default/data.py
```diff
@@ -0,0 +1,77 @@
...
+        APPLICATION_STORE_API_HOST
+        + APPLICATION_ENDPOINT.format(application_id=application_id)
+    )
+    applications_response = get_data(applications_endpoint)
+    return applications_response
+
+
+def extract_data_from_application(application_data, data_subset):
+    data_subset = application_data[data_subset]
+    return data_subset
```
SentientHousePlant, line 47: 🔧 Wouldn't `return application_data[data_subset]` be better? 
🤔 I wouldn't mind a docstring here to explain what this operation is and why it needs to encapsulated by its own function.

app/default/helpers.py
SentientHousePlant, line 5: 🏕️  We are now entering the phase of development where the code we are writing isn't self-explanatory. It would be good to have some docstrings here so future generations understand the context this function lives in.

SentientHousePlant: IE This function posts to the xgov form runner so that ....

This also makes it easier to refactor - since we are documenting the function of the code rather than the implementation.

SentientHousePlant (changes requested):
Nice code! 

When you're back:

- [ ] ☎️ me so we can chat about save / return.

app/default/data.py
Adam-W1: good shout Jack- these helpers are also in the assessor repo so could do with refactoring there too 👍

app/default/data.py
Adam-W1: changed and docstring added - will make it a habbit now :)

app/default/data.py
Adam-W1: changed and added docstrings 👍

app/default/helpers.py
Adam-W1: another good rename suggestion 🥇

app/default/routes.py
Adam-W1: good shout 🔥

PR #37, opened 2022-05-05
FS-597 zap update - FOR REVIEW - DO NOT MERGE YET

harryyo:


Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG BASE_IMAGE_TAG="3.19.0-rc.835"
-FROM ghcr.io/xgovformbuilder/digital-form-builder-runner:$BASE_IMAGE_TAG as base
```
gidsg, line 2: Could this be in the `communitiesuk` org rather than your personal one?

endamccormack: Yeah I agree

endamccormack: Should we not just revert the changes to this file?

endamccormack:
This needs a description @harryyo. It seems wrong to me reading the PR but I dont know why you've done anything.

PR #40, opened 2022-05-23
Fs 880 tasklist

Adam-W1:


app/default/data.py
```diff
@@ -0,0 +1,43 @@
+import json
+import os
+
+import requests
+from app.config import FLASK_ROOT
+from app.config import GET_APPLICATION_ENDPOINT
+
+
+def get_data(endpoint: str):
+    if endpoint[:8] == "https://":
```
SentientHousePlant, line 10: 🤔 Try using [startswith](https://www.programiz.com/python-programming/methods/string/startswith) for better comprehension.

app/default/data.py
```diff
@@ -0,0 +1,43 @@
...
+        data = get_remote_data(endpoint)
+    else:
+        data = get_local_data(endpoint)
+    return data
+
+
+def get_remote_data(endpoint):
```
SentientHousePlant, line 17: 📌 This is a great candidate for monkey patching!

app/models/continue_application.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+        application_id (str): The id of an application in the application store
+        form_name (str): The name of the application sub form/section
+        page_name (str): The form page to redirect the user to.
+
+    Returns:
+        A function: given a users application id they are redirected to
+        the specified application form page within the form runner service
+    """
```
SentientHousePlant, line 24: 💯 These docstrings slay.

app/models/tasklist.py
```diff
@@ -0,0 +1,46 @@
...
+            )
+        ),
+        "number_of_incomplete_sections": len(
+            list(
+                filter(
+                    lambda section: section["status"] == "NOT_STARTED",
+                    application_response["sections"],
+                )
+            )
+        ),
```
SentientHousePlant, line 37: 🔧 I believe that you can do this with a list comprehension...<3

SentientHousePlant (changes requested):
Just some minor comments for you Adam, overall very nice.

app/default/data.py
Adam-W1: oo i do love a new python method

app/models/tasklist.py
Adam-W1: can we pair up on this tomorrow?

PR #41, opened 2022-05-24
FS-877: add landing page

gidsg:
Add dashboard which reads from the application store to get existing applications and POSTs to application store to create new applications 

Note: This reads account ID from a path param which should be replaced once we can read it from the session. Also this assumes there will always be at least one application created, as it relies on params of the existing application to derive the fund and round of the new one (let me know if there's a better way to derive this
Screenshot 2022-05-24 at 14 46 37

### DIFF
--- a/app/default/routes.py
+++ b/app/default/routes.py
@@ -46,6 +46,15 @@ def new(account_id):
             "fund_id": request.form["fund_id"],
         },
     )
+    new_application_json = new_application.json()
+    if new_application.status_code != 201 or not new_application_json.get(
+        "id"
+    ):
+        raise Exception(
+            "Unexpected response from application store when creating new"
+            " application: "
+            + str(new_application.status_code)
+        )
     return redirect(
         url_for(
             "routes.tasklist", application_id=new_application.json().get("id")
--- a/app/default/routes.py
+++ b/app/default/routes.py
@@ -1,7 +1,9 @@
 import requests
+from app.config import APPLICATION_STORE_API_HOST
 from app.config import FORM_REHYDRATION_URL
 from app.config import FORMS_SERVICE_PUBLIC_HOST
 from app.config import SUBMIT_APPLICATION_ENDPOINT
+from app.models.application_summary import ApplicationSummary
 from app.models.continue_application import continue_form_section
 from app.models.eligibility_questions import minimium_money_question_page
 from app.models.tasklist import tasklist_page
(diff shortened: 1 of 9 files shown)


app/models/application_summary.py
```diff
@@ -0,0 +1,34 @@
...
+        )
+
+    @classmethod
+    def from_dict(cls, d: dict):
+        # Filter unknown fields from JSON dictionary
+        return cls(
+            **{
+                k: v
+                for k, v in d.items()
+                if k in inspect.signature(cls).parameters
```
Adam-W1, line 32: really cool class method here - i think we can reuse this elsewhere 👍

Adam-W1 (approved):

app/default/routes.py
```diff
@@ -21,6 +23,34 @@ def index():
...
+        url=f"{APPLICATION_STORE_API_HOST}/applications",
+        json={
+            "account_id": account_id,
+            "round_id": request.form["round_id"],
+            "fund_id": request.form["fund_id"],
+        },
+    )
+    return redirect(
+        url_for("tasklist", application_id=new_application.json().get("id"))
+    )
```
harryyo, line 51: Should there be some validation here (of the form/csrf etc) and of the response (which may be an error message without an id). Perhaps this logic should be a post on the dashboard route and any creation error message could be displayed there?

app/models/application_summary.py
harryyo, line 32: Beautiful!

app/templates/dashboard.html
```diff
@@ -0,0 +1,67 @@
+{%- from "govuk_frontend_jinja/components/tag/macro.html" import govukTag -%}
+{%- from "govuk_frontend_jinja/components/table/macro.html" import govukTable -%}
+{%- from "govuk_frontend_jinja/components/button/macro.html" import govukButton -%}
+{% extends "base.html" %}
+{% block content %}
+    <h1 class="govuk-heading-xl">Your applications</h1>
+    <p class="govuk-body">
+        You have started {% trans count=applications|length %} {{ count }} application {% pluralize %} {{ count }} applications {% endtrans %} using this email address.
```
harryyo, line 8: Very nice

app/templates/dashboard.html
```diff
@@ -0,0 +1,67 @@
...
+    </p>
+    {% set ns = namespace(rows = []) %}
```
harryyo, line 10: Really nicely organised! - Only thought, iis there a linter or something for jinja to handle indentation - sometimes a little harder to read without it

gidsg: In fact I did run it through a linter `djlint` and this is what came out 🤷

gidsg: @harryyo I did run it with that profile and that's what it outputs
```
$ djlint --profile=jinja --reformat app/templates/dashboard.html

Reformatting 1/1 files ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 00:00    


0 files were updated.
```
We can still tweak it if you have specific things you think make it hard to read.

app/default/routes.py
gidsg: As discussed offline, I believe the app is validating the csrf token in the HTML form. (you can validate this by removing it from the HTML form.) We could have more error handling but I think a 500 would be appropriate if something went wrong and currently we don't have a way to usefully log errors which is a broader issue.

app/templates/dashboard.html
```diff
@@ -0,0 +1,67 @@
...
+'text': 'Started on'
+},
+{
+'text': 'Last edited'
+}
+],
+'rows': ns.rows
+}) }}
+<form method="POST" action={{ url_for('routes.new', account_id=account_id) }}>
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
```
gidsg, line 57: @harryyo you can see we are generating the CSRF token here and without it the form POST will fail.

harryyo (approved):
Looks great

PR #44, opened 2022-05-27
FS-880-Workflow

Adam-W1:
code refactoring and workflow updates


.github/workflows/govcloud.yml
```diff
@@ -92,12 +91,56 @@ jobs:
...
+  #       uses: zaproxy/action-baseline@v0.6.1
+  #       with:
+  #         target: ${{ secrets.FORMS_SERVICE_HOST }}
+  #         allow_issue_writing: False
+
+  #     - name: ZAP Full Scan
+  #       uses: zaproxy/action-full-scan@v0.3.0
+  #       with:
+  #         target: ${{ secrets.FORMS_SERVICE_HOST }}
+  #         allow_issue_writing: False
```
gidsg, line 138: was it intentional to comment out this block? can it be either uncommented or removed?

.github/workflows/govcloud.yml
```diff
@@ -109,31 +152,124 @@ jobs:
...
+
+  #     - name: Run performance tests
+  #       run: ls && python -m pip install locust && python -m locust
+  #       working-directory: ./funding-service-design-performance-tests
+  #     - name: "Upload Artifact"
+  #       uses: actions/upload-artifact@v2
+  #       with:
+  #         name: performance-test-report
+  #         path: ./funding-service-design-performance-tests/locust_html_report.html
+  #         retention-days: 5
```
gidsg, line 275: As above, was it intentional to comment out this block? can it be either uncommented or removed?

app/default/routes.py
```diff
@@ -74,18 +78,87 @@ def not_eligible():
...
+
+    Returns:
+        function: a function which renders the tasklist template.
+    """
+    application_response = get_application_data(application_id)
+    application = Application.from_dict(application_response)
+    if not (application and application.sections):
+        return render_template(
+            "application_unknown.html", application_id=application_id
+        )
```
gidsg, line 95: I guess the sections error should never happen by a user's action, so I would suggest throwing a 500 in that case and an exception, if the application's not found, fine to return a 404 response.

.github/workflows/govcloud.yml
Adam-W1: The performance tests are awaiting @rayhan15243 changes—so on pause for now.
The runner testing fails on the 404 error, I believe the fix is underway on this repo.

app/default/routes.py
Adam-W1: agreed – removed unnecessary page and sending to a 404 if application not found 👍

manifest.yml
```diff
@@ -19,3 +36,4 @@ applications:
   sandbox: true
   env:
     PREVIEW_MODE: TRUE
+    NODE_CONFIG: '{"safelist": ["funding-service-design-application-store-dev.london.cloudapps.digital"]}'
```
gidsg, line 39: Do we still need this as we're setting it in the config file now?

.github/workflows/govcloud.yml
gidsg: OK, can we add a TODO then to uncomment this when the tests are fixed?

app/templates/partials/tasklist_section.html
```diff
@@ -9,11 +9,11 @@ <h2 class="app-task-list__section">
       {% for question_page in section["questions"]%}
       <li class="app-task-list__item">
         <span class="app-task-list__task-name">
-          <a href="/continue_application/{{application_id}}?section_name={{section["section_name"]}}&page_name={{question_page["question"]}}" aria-describedby="page-title">
-            <p id="page-title" class="govuk-body">{{question_page["question"]}}</p>
+          <a class="govuk-body govuk-link govuk-link--no-visited-state" href="/continue_application/{{application_id}}?section_name={{section["section_name"]}}&page_name={{question_page["question"]}}">
```
gidsg, line 12: Can we use `url_for` to build this route?

app/models/application.py
```diff
@@ -0,0 +1,36 @@
...
+    id: str
+    account_id: str
+    status: str
+    fund_id: str
+    round_id: str
+    project_name: str
+    date_submitted: datetime
+    started_at: datetime
+    last_edited: datetime
+    sections: List
```
gidsg, line 18: Can we be any more specific about the type within the `List`?

app/templates/partials/tasklist_section.html
Adam-W1: i think we can actually, from the docs: Variable arguments that are unknown to the target endpoint are appended to the generated URL as query arguments. Ill give it a go

build.py
```diff
@@ -50,6 +50,9 @@ def build_assets():
 
     print("Deleting app/static/assets")
 
+    # Copy css
+    shutil.copyfile("static/src/tasklist.css", "app/static/tasklist.css")
+
```
gidsg, line 55: I think we could safely remove this step and just commit the file directly in `app/static/styles`

gidsg: Oh wait I think this is because `app/static` is ignored by .gitignore. I still think `app/static/styles` makes more sense as a path.

gidsg (approved):
🏁 👍

stevewalton28 (approved):

PR #47, opened 2022-06-14
Fs 1083 fix status display

Adam-W1:


SentientHousePlant (approved):
LGTM 🔮

app/filters.py
```diff
@@ -3,8 +3,10 @@ def datetime_format(value, format="%d/%m/%y"):
...
 def snake_case_to_human(word):
-    return word.replace("_", " ").title()
+    if word:
+        return word.replace("_", " ").title()
 
 
 def kebab_case_to_human(word):
-    return word.replace("-", " ").title()
+    if word:
+        return word.replace("-", " ").title()
```
harryyo, line 12: May be worth a comment/docstring on this to note that this shouldn't be used to unslugify as "-" are used to replace other special characters too

harryyo (approved):

app/filters.py
Adam-W1: great shout

PR #51, opened 2022-06-15
Fs 1064 fix submit button

Adam-W1:
Fixed: 
- Submit button is still present for an application after submission
- Removed the completed x of x sections if application submitted
- Removed hardcoded statuses and call from a status set in config


app/config.py
```diff
@@ -53,3 +53,5 @@
 
 DEFAULT_FUND_ID = "funding-service-design"
 DEFAULT_ROUND_ID = "summer"
+
+APPLICATION_STATUSES = ("NOT_STARTED", "IN_PROGRESS", "COMPLETED", "SUBMITTED")
```
gidsg, line 57: Could we use an Enum rather than a tuple for this? This should make the usages much more readable. i.e. https://docs.python.org/3/library/enum.html

rayhan15243 (approved):
@Adam-W1 Tested this in Dev. Looks all good to me.

app/config.py
Adam-W1: Yes that's a great shout - I was a bit worried about the calls not being clear enough 💯

app/default/routes.py
```diff
@@ -1,5 +1,6 @@
 import requests
 from app import config
+from app.config import APPLICATION_STATUSES
```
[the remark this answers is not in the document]
Adam-W1: using Enum as above ^

gidsg:
Did you mean to change permissions on `app/config.py` I think this could be reverted?

app/templates/tasklist.html
```diff
@@ -29,15 +31,16 @@ <h2 class="govuk-body govuk-!-font-weight-bold">
...
   {{ form.csrf_token }}
   <input type="hidden" id="application_id" name="application_id" value="{{application_meta_data["application_id"]}}">
   <button class="govuk-button" data-module="govuk-button"
-  {% if application_meta_data["number_of_incomplete_sections"] == application_meta_data["number_of_sections"] %}
-  disabled
- {% endif %}
+  {% if not application_response["status"] == application_meta_data["completed_status"] %}
+    disabled
+  {% endif %}
   >Submit</button>
```
gidsg, line 43: Could we put a line break after `>` ?

gidsg (approved):
LGTM 👍

SentientHousePlant (approved):

harryyo (approved):
Great

PR #53, opened 2022-06-30
FS-657: add app dockerfile

gidsg:
Adds dockerfile for app 
*(note alternative filename is because `Dockerfile` in this repo was already used for the form-builder and I didn't want to break anything expecting that to exist.)*

Also fixes config to make  local data configurable to make it more usable in docker context.

### DIFF
--- a/config/envs/development.py
+++ b/config/envs/development.py
@@ -1,7 +1,9 @@
 """Flask Local Development Environment Configuration."""
 import logging
+from os import getenv
 
 from config.envs.default import DefaultConfig
+from distutils.util import strtobool
 from fsd_utils import configclass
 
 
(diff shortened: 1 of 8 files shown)


config/envs/development.py
```diff
@@ -1,11 +1,14 @@
...
 from config.envs.default import DefaultConfig
+from distutils.util import strtobool
 from fsd_utils import configclass
 
 
 @configclass
 class DevelopmentConfig(DefaultConfig):
 
     FSD_LOGGING_LEVEL = logging.DEBUG
+    USE_LOCAL_DATA = strtobool(getenv("USE_LOCAL_DATA", "True"))
```
Adam-W1, line 14: I think if one sets USE_LOCAL_DATA in the ENV this will automatically take precedence anyway given the config util

gidsg: Doesn't appear so from my testing, e.g. if I hardcode this to "True" the ENV var is ignored.

Adam-W1: apologies - I got it the wrong way around then, class config overrides the environment

Adam-W1 (approved):

PR #60, opened 2022-07-21
Fs 1190 status fix

donshellywork:
https://digital.dclg.gov.uk/jira/browse/FS-1190

### DIFF
--- a/app/templates/tasklist.html
+++ b/app/templates/tasklist.html
@@ -19,19 +19,20 @@ from "partials/tasklist_section.html" import tasklist_section %} {% block conten
 
     <p class="govuk-body-l govuk-!-font-weight-bold">
       Status: 
-      {% set completed_count = application.question_status_summary.get("COMPLETED") %}
+      {% set completed_count = application_meta_data["number_of_completed_sections"] %}
+      {% set section_count = application_meta_data["number_of_sections"] %}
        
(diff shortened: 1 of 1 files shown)


gidsg:
Can you add a screenshot where we are making a visual change like this?

donshellywork:
@gidsg I couldn't figure out how to nav to the page, i was advised to push anyway

app/templates/tasklist.html
```diff
@@ -9,13 +9,28 @@ <h1 class="govuk-heading-xl">
...
+      <strong class="govuk-tag govuk-tag--blue">
+        In Progress
+      </strong>
       {% endif %}
-    </h2>
+      {% if completed_count == section_count %}
+      <strong class="govuk-tag govuk-tag--green">
+        Complete
+      </strong>
+      {% endif %}
```
gidsg, line 31: Could we use `elif` here as these should be exclusive? e.g. https://jinja.palletsprojects.com/en/3.0.x/templates/#if

donshellywork:
Screenshot 2022-07-26 113109
Screenshot 2022-07-26 112401

app/templates/tasklist.html
donshellywork: good point - implementing

gidsg (approved):
LGTM 👍

PR #62, opened 2022-07-21
fixed title

donshellywork:
https://digital.dclg.gov.uk/jira/browse/FS-1195


gidsg:
Can we update the page `<title>` too? Currently it appears to be `DLUHC Funding Service Design Iteration 2` We need a way to make this unique per page, so probably should be an attribute of the base template we can add a crumbtrail too.

donshellywork:
> 

Will naming the page in base.html "Funding Service Design - Task List" suffice?

tests/conftest.py
```diff
@@ -8,7 +8,7 @@
 from selenium.webdriver.chrome.service import Service
 from webdriver_manager.chrome import ChromeDriverManager
 
-multiprocessing.set_start_method("fork")  # Required on macOSX
+# multiprocessing.set_start_method("fork")  # Required on macOSX
```
gidsg, line 11: Rather than commenting this out can we make it conditional as per: https://github.com/communitiesuk/funding-service-design-notification/blob/703a2eb907d4012671521d8ad47a306ea01f6f77/tests/conftest.py#L13-L14

gidsg:
> Will naming the page in base.html "Funding Service Design - Task List" suffice?

@donshellywork No, as that will apply to all pages. I suspect we will need to use a pattern similar to this to override it on this page only:
https://github.com/communitiesuk/funding-service-design-frontend/blob/54a0faf287be75467d42f01efd23a562a37c991c/app/templates/404.html#L3

app/create_app.py
```diff
@@ -57,11 +57,11 @@ def create_app() -> Flask:
     def inject_global_constants():
         return dict(
             stage="beta",
-            service_title="DLUHC Funding Service Design Iteration 2",
+            service_title="Task List - Funding Service Design",
```
gidsg, line 60:
```suggestion
            service_title="Funding Service Design",
```

app/create_app.py
```diff
@@ -57,11 +57,11 @@ def create_app() -> Flask:
...
             service_meta_description=(
-                "DLUHC Funding Service Design Iteration 2"
+                "Task List - Funding Service Design"
```
gidsg, line 62:
```suggestion
                "Funding Service Design"
```

app/templates/tasklist.html
```diff
@@ -2,11 +2,13 @@
 "govuk_frontend_jinja/components/button/macro.html" import govukButton -%} {%
 from "partials/tasklist_section.html" import tasklist_section %} {% block content %}
 
+{% block pageTitle %}{{ service_title }}{% endblock %}
```
gidsg, line 5:
```suggestion
{% block pageTitle %}Task List - {{ service_title }}{% endblock %}
```
We want to prefix the service name like this.

app/create_app.py
```diff
@@ -57,9 +57,9 @@ def create_app() -> Flask:
     def inject_global_constants():
         return dict(
             stage="beta",
-            service_title="Task List - Funding Service Design",
+            service_title="Funding Service Design",
             service_meta_description=(
-                "Task List - Funding Service Design"
+                "Funding Service Design"
             ),
             service_meta_keywords="Task List - Funding Service Design",
```
gidsg, line 64:
```suggestion
            service_meta_keywords="Funding Service Design",
```

app/create_app.py
```diff
@@ -57,11 +57,11 @@ def create_app() -> Flask:
...
         return dict(
             stage="beta",
-            service_title="DLUHC Funding Service Design Iteration 2",
+            service_title="Funding Service Design",
             service_meta_description=(
-                "DLUHC Funding Service Design Iteration 2"
+                "Funding Service Design"
             ),
-            service_meta_keywords="DLUHC Funding Service Design Iteration 1",
+            service_meta_keywords="Task List - Funding Service Design",
```
gidsg, line 64: Shouldn't have Task List globally here
```suggestion
            service_meta_keywords="Funding Service Design",
```

gidsg (approved):
LGTM 👍

PR #68, opened 2022-07-27
✨  added  about-your-organisation json

ramsharma-prog:
New json files are  added as per the prototype.


gidsg:
What is the context here? Is this a bug fix or just updating data based on new designs?

rayhan15243 (approved):
Checked this section against the prototype as per our call @ramsharma-prog. 

Looks good to me.

ramsharma-prog:
> What is the context here? Is this a bug fix or just updating data based on new designs?

just updating data based on new designs

gidsg (approved):

manifest.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: funding-service-design-form-runner
+  docker:
+    image: ghcr.io/communitiesuk/digital-form-builder/runner:latest
+  memory: 528M
+  sandbox: true
+  routes:
+    - route: forms.dev.fundingservice.co.uk
+  env:
+    PREVIEW_MODE: TRUE
+
```
gidsg, line 12: Can we rebase instead? This file is already on `main` branch.

ramsharma-prog: Thanks Gideon for helping with rebasing 🌟

PR #70, opened 2022-07-28
FS-1163 and FS-1050

srh-sloan:
Changes to the frontend app and forms deploy to support custom domains:
- Added the custom routes to the frontend app for dev and test
- Updated workflow to deploy a separate form runner instance to each of dev and test so support the custom cookie domain

### DIFF
--- a/.github/workflows/govcloud.yml
+++ b/.github/workflows/govcloud.yml
@@ -81,7 +81,7 @@ jobs:
           sudo apt-get install cf8-cli
 
       - name: Deploy to Gov.uk PaaS Dev
-        id: deploy-to-paas
+        id: deploy-to-paas Dev
         shell: bash
         run: |
           cf api ${{ secrets.CF_API }}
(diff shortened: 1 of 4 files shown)


.github/workflows/govcloud.yml
```diff
@@ -216,7 +227,7 @@ jobs:
           allow_issue_writing: False
 
   deploy_test:
-    needs: security
+    #needs: security
```
gidsg, line 230: Did we mean to comment this out?

srh-sloan: Nope, pressed undo one too many times!

gidsg (approved):
LGTM 👍

app/templates/tasklist.html
```diff
@@ -9,13 +9,26 @@ <h1 class="govuk-heading-xl">
...
+      </strong>
+      {% elif completed_count == section_count %}
+      <strong class="govuk-tag govuk-tag--green">
+        Complete
+      </strong>
       {% endif %}
-    </h2>
+    </p>
+    
     <p class="govuk-body govuk-!-margin-bottom-7">
```
gidsg, line 32: This seems unrelated, was it meant to be included?

srh-sloan: I  didn't change this file - just done a rebase before I merged

PR #75, opened 2022-08-01
FS-1196 implement new submit application pattern

stevewalton28:
In preparation for FS-1196 (which will be played in Sprint 19).  We need to implement a non-standard "alert" pattern as per the design in the prototype (here https://fsd-funding-prototype-v2.herokuapp.com/cof-w2/app-task-list-new) when an application is ready to submit.

This submit alert will only be displayed once all sections are complete (see detail for rules in FS-1196).

This initial PR implements the styles and layout in a Jinja component ready to use.

Tasks to complete when picking this up:

1. Remove "example_tasklist_submit.html" from templates and the associated route handled in "example_submit_pattern()"
2. Correctly route the action taken when click "Submit your application" (may require additional incoming variables to the macro

Changes made:

- Temp route added to show the pattern working handled by "example_submit_pattern()" in routes and templated in "example_tasklist_submit.html"
- Added new macro for "tasklist_submit.html"


Adam-W1:
changes now on branch  #94

PR #76, opened 2022-08-01
Fs 1263 publish fork

srh-sloan:
Updated to take the base image as the communitiesuk version of the form-builder: https://github.com/communitiesuk/digital-form-builder/pkgs/container/digital-form-builder-dluhc-runner

### DIFF
--- a/.github/workflows/build-deploy-forms.yml
+++ b/.github/workflows/build-deploy-forms.yml
@@ -1,0 +1,45 @@
+name: Build and deploy forms
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - "form_jsons/**"
+env:
+  DOCKER_REGISTRY: ghcr.io
+jobs:
+  setup-env-vars:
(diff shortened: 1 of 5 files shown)


.github/workflows/xforms_builder.yml
```diff
@@ -2,7 +2,7 @@ name: Build xforms image
 
 on:
   workflow_dispatch:
-  push:
+ # push:
     # branches:
     # - master
     # - FS-389
```
gidsg, line 8: Could we just delete this file? It will still be in git history so could be restored if it was found to be needed?

.github/workflows/build-deploy-forms.yml
```diff
@@ -87,7 +87,7 @@ jobs:
       - name: Run tests
         env:
           # TODO Add in basic auth creds to URL when secret setup
```
gidsg, line 89: we could remove this comment now!!

PR #77, opened 2022-08-03
FS-1263: Make form host public and point to domain

gidsg:
Having this as a secret obfuscated the issues previously


srh-sloan (approved):
Much clearer

PR #78, opened 2022-08-03
fs-1277 Removed bug in jinja

donshellywork:
this leads from fs-1195 where a bug was introduced

https://digital.dclg.gov.uk/jira/browse/FS-1277


.github/workflows/govcloud.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           CF_DOCKER_PASSWORD=${{ secrets.GHCR_PAT }} cf push funding-service-design-form-runner-dev -f ./manifest.yml \
             --docker-image ${{ env.DOCKER_IMAGE_TAG }} \
             --docker-username ${{ github.actor }}
-
```
gidsg, line 93: can we revert these whitespace only changes?

manifest.yml
```diff
@@ -9,7 +9,7 @@ applications:
     - route: forms.dev.fundingservice.co.uk
   env:
     PREVIEW_MODE: TRUE
-
```
gidsg, line 12: can you revert these whitespace only changes?

.github/workflows/govcloud.yml
gidsg: Please revert to the version of this file from the main branch e.g. `git checkout main -- .github/workflows/govcloud.yml`

gidsg (approved):
LGTM 👍

SentientHousePlant (approved):

PR #79, opened 2022-08-03
FS-1263: bind form runner to s3

gidsg:
Previously, deployment of our fork was failing as it was missing the S3 binding e.g.

```
  2022-08-01T17:11:53.31+0100 [APP/PROC/WEB/0] ERR /usr/src/app/runner/dist/server/services/uploadService.js:22
  2022-08-01T17:11:53.31+0100 [APP/PROC/WEB/0] ERR const s3Credentials = vcapJson["aws-s3-bucket"][0].credentials;
  2022-08-01T17:11:53.31+0100 [APP/PROC/WEB/0] ERR ^
  2022-08-01T17:11:53.31+0100 [APP/PROC/WEB/0] ERR TypeError: Cannot read property '0' of undefined
```
This should fix that error by binding the apps to the bucket

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -9,7 +9,9 @@ applications:
     - route: forms.dev.fundingservice.co.uk
   env:
     PREVIEW_MODE: TRUE
-    
+  services:
+    - form-uploads-dev
+
 - name: funding-service-design-form-runner-test
   docker:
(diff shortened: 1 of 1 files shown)


srh-sloan (approved):
As discussed, need to make sure we remember to configure these services in future envs

PR #80, opened 2022-08-03
Add redirect option

adamdavies1:


app/default/routes.py
```diff
@@ -193,4 +198,4 @@ def not_found(error):
 
 @default_bp.errorhandler(500)
 def internal_server_error(error):
-    return render_template("500.html"), 500
+    return render_template("500.html"), 500
```
adamdavies1, line 201: No sure why this is showing as a change, now new line breaks added, no spaces.

gidsg: It is because there is no newline character at the end.

gidsg (approved):
LGTM 👍 (you'll need to mark as not-draft to merge)

endamccormack (approved):

PR #88, opened 2022-08-08
FS-855 Configuring frontend for COF fund and round

stevewalton28:


app/default/routes.py
```diff
@@ -65,11 +68,12 @@ def new(account_id):
         url=f"{Config.APPLICATION_STORE_API_HOST}/applications",
         json={
             "account_id": account_id,
-            "round_id": request.form["round_id"] or Config.DEFAULT_ROUND_ID,
-            "fund_id": request.form["fund_id"] or Config.DEFAULT_FUND_ID,
+            "round_id": Config.DEFAULT_ROUND_ID,
+            "fund_id": Config.DEFAULT_FUND_ID,
```
gidsg, line 72: Should we keep the previous fallback, which was designed to allow multiple funds passed by query string?

app/templates/tasklist.html
```diff
@@ -9,8 +9,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">
-      {{application_meta_data["fund"]|kebab_case_to_human}}
-      application
+      Community ownership fund application
```
gidsg, line 12: Again could we retain the previous logic to make this more generic?

app/default/routes.py
```diff
@@ -106,13 +111,20 @@ def tasklist(application_id):
         Config.GET_APPLICATION_ENDPOINT.format(application_id=application_id)
     )
     if not (application_response and application_response["sections"]):
-        return abort(404)
+        return abort(500)
```
gidsg, line 114: Can we log something before we abort about what went wrong?

stevewalton28: Yeh great idea I'll put that in

gidsg (approved):
LGTM ⛽ 👍

PR #96, opened 2022-08-12
Fs 1288 refactor with new application store

Adam-W1:
The tasklist design requires changes to how we handle sections in the application store and subsequently the task list.
The hyperlinks under the headings in the task list used to be pages in a single json, now each hyperlink under a heading represents an individual form.

Update application store to handle puts to a form instead of section (a section is now a frontend/visual concept), an application is associated with multiple forms.
This PR includes the configuration and implementation of the section concept as reflected in the prototype designs.


config/envs/default.py
```diff
@@ -127,3 +133,60 @@ class DefaultConfig:
     }
 
     USE_LOCAL_DATA = strtobool(getenv("USE_LOCAL_DATA", "False"))
+
+    COF_FUND_ID = "47aef2f5-3fcb-4d45-acb5-f0152b5f03c4"
+    COF_R2_ROUND_ID = "c603d114-5364-4474-a0c4-c41cbf4d3bbd"
```
gidsg, line 138: Doesn't seem like these are being used at all?

config/envs/default.py
```diff
@@ -127,3 +133,60 @@ class DefaultConfig:
...
+    COF_R2_SECTION_CONFIG = {
```
gidsg, line 139: I think this is too big to be in Config, can we put it in a more context specific class?
Is there a reason this doesn't exist in a store?

Adam-W1: As part of the conversations discussing the new db schema (reflective of the design and form json changes), it was brought up that the presentation layer (the sections) are a frontend concept (and will also be different in the assessment frontend), the application store holds all the forms associated with an application ID but does not enforce a presentation structure. 
We now configure section/form order in the frontend service using a predefined config json as above. In the longer term it might make more sense to have some mappings from the fund/round store to the application store once we have details on how we want to structure the fund data but this information does not yet exist.

gidsg: OK, can we move this into a related class in that case rather than directly in Config? I don't think this is something we'd want to be overriding on a per environment basis?

Adam-W1: sure - that's a good suggestion 💯

config/envs/default.py
Adam-W1: true, ill remove. We need them to get fund info from the fund store (https://github.com/communitiesuk/funding-service-design-frontend/blob/fs-1288-refactor-with-new-application-store/app/models/application.py#:~:text=%5D%20%3D%20form-,%23%20TODO%20post%20mvp,-%23%20Update%20the%20section) 
...however the fund_id and round_id are available on the application object 👍

tests/test_application_form.py
```diff
@@ -101,8 +101,9 @@ def test_form_pages_accessible(self):
                     },
                     route_rel,
                 )
-                assert len(results["violations"]) <= 4
-                assert len(results["violations"]) <= 4 or all(
+                # TODO why is this number 5?
```
harryyo, line 104: Violations I think should be reported in axe_reports when you run locally?

app/templates/tasklist.html
```diff
@@ -11,58 +12,44 @@
...
+    {{tasklist_submit(application_id, application_meta_data, form, url_for('routes.submit_application'))}}
+    {% elif not application["status"] == application_meta_data["submitted_status"] %}
+    <h2 class="govuk-body-l govuk-!-font-weight-bold">
       Status:
-      {% set completed_count = application_meta_data["number_of_completed_sections"] %}
-      {% set section_count = application_meta_data["number_of_sections"] %}
-      {% if completed_count == 0 %}
+      {% set completed_count = application_meta_data["number_of_completed_forms"] %}
+      {% set in_progress_count = application_meta_data["number_of_in_progress_forms"] %}
+      {% set form_count = application_meta_data["number_of_forms"] %}
```
harryyo, line 23: Are these used?

harryyo (approved):
Minor comments but looks great! 👏

app/templates/tasklist.html
Adam-W1: yes they are used to display counts in tasklist.html

tests/test_application_form.py
Adam-W1: ah nice - thanks Harry!

gidsg (approved):
LGTM 👍

PR #99, opened 2022-08-17
FS 1341 update application forms

ramsharma-prog:


form_jsons/public/applicant-information.json
```diff
@@ -85,7 +85,16 @@
...
   "conditions": [],
   "fees": [],
-  "outputs": [],
+  "outputs": [
+    {
+      "name": "update-form",
+      "title": "Update form in application store",
+      "type": "savePerPage",
+      "outputConfiguration": {
+        "savePerPageUrl": "{{application_store_host}}/applications/forms"
```
Adam-W1, line 94: I think we need to put the host (application_store_host}} into the JSON dynamically

Adam-W1: so the problem is:
we COPY the [forms jsons ](https://github.com/communitiesuk/funding-service-design-frontend/tree/main/form_jsons/public) onto the form runner container as part of the [DOCKER file workflow](https://github.com/communitiesuk/funding-service-design-frontend/blob/main/Dockerfile#:~:text=COPY%20./form_jsons/public,config/*%20runner/config/). 
However the form jsons r[eferences the application store](https://github.com/communitiesuk/funding-service-design-frontend/blob/FS-1341-update-application-forms/form_jsons/public/applicant-information.json#:~:text=%22savePerPageUrl%22%3A%20%22%7B%7Bapplication_store_host%7D%7D/applications/forms%22) as an output (for safe per page functionality) . 
We need to load in the correct application store host ({{application_store_host}}=127.0.0.1:5001 locally) into the forms JSON callbackUrl (for the correct env), so when they are copied into the runner, it PUTs back to the correct application store endpoint. 

 Could this be done by passing through the application store domain as part of the build form runner fork workflow?

gidsg: @Adam-W1 as discussed, I think ideally this is best passed-in at runtime as we don't want to build multiple versions of the form runner image per-environment.

Adam-W1 (approved):

gio-karunakaran (approved):
LGTM :)

PR #102, opened 2022-08-19
FS-987: update safelist with production host

gidsg:
This is needed so the production form runner can communicate with the production application store.


form_jsons/config/development.json
```diff
@@ -3,7 +3,7 @@
     "previewMode": true,
     "enforceCsrf": false,
     "env": "development",
-    "safelist": ["funding-service-design-application-store-dev.london.cloudapps.digital", "funding-service-design-application-store-test.london.cloudapps.digital"],
+    "safelist": ["funding-service-design-application-store-dev.london.cloudapps.digital", "funding-service-design-application-store-test.london.cloudapps.digital", "funding-service-design-application-store.apps.internal"],
```
srh-sloan, line 6: Should we list the prod host in the dev config? Not having it would give an extra layer of protection against hitting prod by mistake from dev

srh-sloan (commented):
Just one comment on the config

form_jsons/config/development.json
gidsg: Fair point @srh-sloan although, I'm not sure currently we're actually using the development config on our test environments (but that's a separate issue). I've reverted this change and force-pushed.

srh-sloan (approved):
🚢 👍

PR #104, opened 2022-08-22
FS-1330-start-application-button

ramsharma-prog:
very tiny PR to add the **secondary govuk button** if there an active application else **green govuk button**
Screenshot 2022-08-30 at 16 58 28


adamdavies1 (commented):
LGTM

app/templates/dashboard.html
```diff
@@ -69,7 +69,20 @@ <h1 class="govuk-heading-xl">Your applications</h1>
...
+
+{% if applications|length > 0 %}
+<form method="POST" action={{url_for('routes.new')}}>
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+    <input type="hidden" name="account_id" value="{{ account_id }}">
+    <input type="hidden" name="fund_id" value="{{ fund_id }}">
+    <input type="hidden" name="round_id" value="{{ round_id }}">
+    {{ govukButton({
+    'text': "Start new application",
+    'classes': "govuk-button--secondary"
```
harryyo, line 81: Ah I think it may be possible to add a conditional after "govuk-button--secondary" like `"govuk-button--secondary" if applications|length > 0` - does it work? If that's the only requirement.

harryyo (commented):
Just a thought - I think you can conditionals in macros maybe?

app/templates/dashboard.html
ramsharma-prog: It didn't work for me 🤔.  Also was unable to call the govuk css class  to the macro so had to create macro for both buttons.

gio-karunakaran (approved):
Looks Great Ram 👍🏾

app/templates/dashboard.html
```diff
@@ -69,13 +71,8 @@ <h1 class="govuk-heading-xl">Your applications</h1>
...
-    <input type="hidden" name="round_id" value="{{ round_id }}">
-    {{ govukButton({
-    'text': "Start new application"
-    }) }}
-</form>
+
+{{startApplicationButton(form, url_for('routes.new'), applications=applications)}}
+
 {% endblock %}
+
```
harryyo, line 78: Is this a spare line or space or something?

harryyo (approved):
Just a new line with a space needs removing, part from that all good!

app/templates/dashboard.html
ramsharma-prog: I ran pre commit hook on the whole service by mistake 😆

PR #106, opened 2022-08-23
FS-1323 & FS-1322 add jwt session expiry timeout and redirect

harryyo:
### Secure tasklist route and increase session timeout on form runner

- @login_required added to tasklist route
- form runner config jsons updated with jwtRedirectToAutheticationUrl which points the runner to the new authenticator sign-out endpoint
- form runner config jsons updated with sessionTimeout of 1 day in milliseconds to push session to last beyond the jwt cookie expiry timeout of one day


form_jsons/config/development.json
```diff
@@ -5,5 +5,6 @@
     "env": "development",
     "safelist": ["funding-service-design-application-store-dev.london.cloudapps.digital", "funding-service-design-application-store-test.london.cloudapps.digital"],
     "jwtAuthCookieName": "fsd_user_token",
-    "jwtAuthenticationUrl": "https://authenticator.dev.fundingservice.co.uk/service/magic-links/invalid"
+    "jwtAuthenticationUrl": "https://authenticator.dev.fundingservice.co.uk/service/magic-links/invalid",
```
harryyo, line 8: This url will need to be updated obviously when we have the /signout (signed out) route available on the authenticator

Adam-W1 (approved):
nice 👌

form_jsons/config/production.json
```diff
@@ -4,5 +4,6 @@
     "enableCsrf": true,
     "safelist": ["funding-service-design-application-store-dev.london.cloudapps.digital", "funding-service-design-application-store-test.london.cloudapps.digital", "funding-service-design-application-store.apps.internal"],
     "jwtAuthCookieName": "fsd_user_token",
-    "jwtAuthenticationUrl": "https://authenticator.live.fundingservice.co.uk/service/magic-links/invalid"
+    "jwtRedirectToAuthenticationUrl": "https://authenticator2.test.gids.dev/sessions/sign-out",
```
gidsg, line 7: As this will vary by environment, I guess we'll need to use an env var in the manifest, rather than baking it in the docker image.

harryyo: Yep good point, was wondering the best way to handle this. We do have a pattern of copying in these config .json files (for development and production at the moment) which are then switched by the global NODE_ENV env var. Should we just make more of these config .json files? Might be an easier way of managing these non-secret settings than having to set lots of env vars in every environment?

gidsg: @harryyo You can set it in the manifest as a regular ENV VAR as per: https://github.com/communitiesuk/digital-form-builder/blob/09986450215c16cd612aae5a5d575177789a2a01/runner/config/custom-environment-variables.json#L48 so no need to rely on `NODE_ENV`

gidsg: @harryyo I'd prefer more in ENV VARs as a rule as it keeps the app more 12-factor-y.

harryyo: @gidsg so you reckon we move all these config vars into the manifest-form-runner.yml and remove this config copying in? Did you manage to get the safelist working with an env var too @Adam-W1 ?

manifest-form-runner.yml
```diff
@@ -8,9 +8,11 @@ applications:
   routes:
     - route: forms.dev.gids.dev
   env:
-    PREVIEW_MODE: TRUE
+    NODE_ENV: production
     RSA256_PUBLIC_KEY_BASE64: ((RSA256_PUBLIC_KEY_BASE64))
     SERVICE_START_PAGE: https://frontend.dev.gids.dev/account
+    JWT_REDIRECT_TO_AUTHENTICATION_URL: https://authenticator2.dev.gids.dev/sessions/sign-out
+    SAFELIST: '["funding-service-design-application-store-dev.london.cloudapps.digital"]'
```
gidsg, line 15: Don't think this will work as it's not defined here: https://github.com/communitiesuk/digital-form-builder/blob/main/runner/config/custom-environment-variables.json ?
You might need to leave it in `NODE_CONFIG` for now.

form_jsons/config/production.json
gidsg: @harryyo I am happy for things that don't vary by environment to still be in the config that gets copied in.

SentientHousePlant (approved):

form_jsons/config/production.json
harryyo: Done

manifest-form-runner.yml
harryyo: Reverted to having this defined in copied in [production.json](https://github.com/communitiesuk/funding-service-design-frontend/pull/106/files#diff-45ba9790f0e88dec4753d77532166925ad4337c545d91b90db6a429b779d53c7)

PR #113, opened 2022-08-31
FS-1410 Remove application form runner tests

harryyo:
### Remove test_application_form tests from test set and workflow jobs.

Following a discussion with @stevewalton28 @Adam-W1 @rayhan15243 it was decided that these content tests do not need to (and strictly should not) be performed by the frontend, and should instead be moved to full user centric end-to-end test performed by our end to end testing suite led by @rayhan15243.


Adam-W1 (approved):
Looks good 👍

ramsharma-prog (approved):

PR #118, opened 2022-09-02
Update start page content

AbuNavsa:
Updated start page content to match design, dummy/blank links where page has not been created yet


Adam-W1 (approved):
LGTM 👍

PR #123, opened 2022-09-10
Updating content for COF launch

stevewalton28:
Several changes included here:

1. Pull contact details from fund stores round
2. Add contact us page to footer
3. Create partial for contact us details, pulling in phone and text phone if they are set in the round object in the fund store (so we can quickly add these if required close to launch)
4. Added logging to 500 and 404 (logging incoming error object)
5. Updating start page to match prototype
6. Updating tasklist to match prototype
7. Updating submission date content to pull from fund/round store - to update this now, we simply need to make a small PR to the fund store instead of making lots of content changes


app/models/round.py
```diff
@@ -1,18 +1,55 @@
...
+        # Filter unknown fields from JSON dictionary
+        return cls(
+            **{
+                k: v
+                for k, v in d.items()
+                if k in inspect.signature(cls).parameters
+            }
+        )
+
+
```
harryyo, line 41: Are these needed if we're just using these as dicts? SupportAvailablity has a "closed" attribute too currently (from fund_store) so maybe should either remove these and change the contact_details support_avaialability types to dict (Lines 51/52 below). Or have them explicitly match what's coming from fund_store and instantiate them as ContactDetails/SupportAvailability objects when creating the Round object from_dict()?

harryyo (commented):
Ah a lot of work here - could use the ContactDetails/SupportAvailability classes or remove if just using them as dicts?

app/default/routes.py
```diff
@@ -23,15 +25,44 @@
...
 def accessibility_statement():
     current_app.logger.info("Accessibility statement page loaded.")
-    return render_template("accessibility-statement.html")
+    return render_template("accessibility_statement.html")
+
+
+@default_bp.route("/cof-r2w2-all-questions", methods=["GET"])
+def all_questions():
+    current_app.logger.info("All questions page loaded.")
+    round_data = get_round_data(Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID, as_dict=True)
```
srh-sloan, line 47: A bit nitpicky... but do we need the network call here? If the whole page is hardcoded questions, and called `cof-r2w2-etc` can we just hardcode the round name too? Then this is just static content and we don't need to `render_template`

app/templates/404.html
```diff
@@ -1,13 +1,17 @@
...
-    <p class="govuk-body">If you typed the web address, check it is correct.</p>
-    <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
-  </div>
-</div>
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
+            <p class="govuk-body">Try again later.</p>
+            <p class="govuk-body">If the web address is correct, or you selected a link or button, contact the Community Ownership Fund team
+                to speak to an adviser. </p>
```
srh-sloan, line 13: Same as authenticator - you might not be able to speak to an advisor

srh-sloan (commented):
A few comments, including some content ones.

app/default/routes.py
stevewalton28: Yep, good spot! Fixed.

app/default/routes.py
```diff
@@ -23,15 +25,44 @@
...
+    return render_template(
+        "cof-r2w2-all-questions.html",
+        round_title=round_data.title,)
+
+
+@default_bp.route("/contact_us", methods=["GET"])
+def contact_us():
+    current_app.logger.info("Contact us page loaded.")
+    round_data = get_round_data(Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID, as_dict=True)
```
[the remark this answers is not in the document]
stevewalton28: Fixed

app/templates/tasklist.html
```diff
@@ -3,55 +3,130 @@
...
+
+                    <h2 class="govuk-heading govuk-heading-s">
+                        Submitting your application
+                    </h2>
+                    <p class="govuk-body">
+                        When you have completed all sections, you'll be able to submit your application.
+                    </p>
+                    <p class="govuk-body">
+                        Once you have submitted your application, you'll not be able to edit, add or change any
+                        sections.
```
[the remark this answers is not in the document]
stevewalton28: This is the designed content - have asked, if needed iterations made post launch

app/templates/500.html
```diff
@@ -9,10 +9,7 @@
         <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
             <p class="govuk-body">Try again later.</p>
-            <p class="govuk-body">We have not saved your answers. When the service is available, you will have to start
-                again. </p>
-            <p class="govuk-body">Contact the Community Ownership Fund if you have any questions. </p>
-            {{ contact_details(contact_us_email_address, contact_us_phone, contact_us_text_phone, opening_days, opening_time, closed) }}
+            <p class="govuk-body"><a class="govuk-link" href="{{ contact_us_url }}">Contact the Community Ownership Fund</a> if you have any questions.</p>
```
srh-sloan, line 12: Where is `contact_us_url` being populated?

app/templates/cof_r2w2_all_questions.html
```diff
@@ -7,7 +7,11 @@
 {% block content %}
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-        <span class="govuk-caption-l">Community Ownership Fund {{ round_title }}</span>
+            {% if round_data.title|length %}
```
srh-sloan, line 10: Why are we checking the length if we don't then use `round_data.title`?

form_jsons/public/value-to-the-community.json
```diff
@@ -85,10 +85,10 @@
   ],
   "version": 2,
   "skipSummary": false,
-  "name": "Apply for funding to save a building in your community",
+  "name": "Apply for funding to save an asset in your community",
```
srh-sloan, line 88: Not now, but for future funds we should really find a way to store this text in just one place!

app/default/data.py
```diff
@@ -103,3 +103,20 @@ def get_round_data(fund_id, round_id, as_dict=False):
...
         return round_response
+
+
+def get_round_data_fail_gracefully(fund_id, round_id):
+    try:
+        round_request_url = Config.GET_ROUND_DATA_FOR_FUND_ENDPOINT.format(
+            fund_id=fund_id, round_id=round_id
+        )
+        round_response = get_data(round_request_url)
+        current_app.logger.error(f"{ round_response }")
```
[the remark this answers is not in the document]
stevewalton28: It is an error if we make a call to the fund store and it fails - we're logging as an error because it should never fail - we are tolerant of its failure, but if its down we need to know.

app/templates/500.html
```diff
@@ -9,10 +9,7 @@
         <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
             <p class="govuk-body">Try again later.</p>
-            <p class="govuk-body">We have not saved your answers. When the service is available, you will have to start
-                again. </p>
-            <p class="govuk-body">Contact the Community Ownership Fund if you have any questions. </p>
-            {{ contact_details(contact_us_email_address, contact_us_phone, contact_us_text_phone, opening_days, opening_time, closed) }}
```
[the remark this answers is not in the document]
stevewalton28: Done

app/templates/500.html
stevewalton28: Done

app/templates/cof_r2w2_all_questions.html
stevewalton28: Because we're being tolerant of the round call failing and the round_data being empty - if its empty, we want to fail gracefully so we check if the title is set - if it isn't we just dont render that content rather than bring the whole service down

form_jsons/public/value-to-the-community.json
stevewalton28: Agreed!

app/default/data.py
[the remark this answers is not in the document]
srh-sloan: but isn't the failure condition what's in the `except` block below? This line looks like part of the 'it worked' path?

app/templates/500.html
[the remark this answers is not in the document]
srh-sloan: It's still there?

app/templates/cof_r2w2_all_questions.html
srh-sloan: But why not just hard code `Community Ownership Fund Round 2 Window 2` regardless of whether title is there? It's hardcoded if it's there (line 11)

app/models/round.py
stevewalton28: Will fix in future PR

app/default/data.py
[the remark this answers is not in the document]
stevewalton28: Aye I was looking at the wrong line your right I've removed .

app/templates/cof_r2w2_all_questions.html
stevewalton28: Your dead right, I'd only fixed half of it after our chat before.

srh-sloan (approved):
Good to go 👍

PR #134, opened 2022-09-20
Fs 1516 fix org type

srh-sloan:
Updated the different options for `organisation-type` to have unique values, previously they had the same values. Also updated the `conditions` section to account for these unique values.

During investigating realised there were some duplicate, unused conditions around this field so I have removed these.

### DIFF
--- a/form_jsons/public/organisation-information.json
+++ b/form_jsons/public/organisation-information.json
@@ -595,19 +595,28 @@
         "items": [
           {
             "text": "Charitable incorporated organisation (CIO)",
-            "value": "CHN"
+            "value": "CIO"
           },
           {
             "text": "Co-operative, such as a community benefit society",
-            "value": "RGN"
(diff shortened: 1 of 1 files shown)


form_jsons/public/organisation-information.json
```diff
@@ -839,10 +830,10 @@
         }
       },
       {
-        "displayName": "Co-operative, such as a community benefit society",
-        "name": "MomaFn",
+        "displayName": "Type of organisation-  option(two, three, four)",
+        "name": "mNExjy",
```
gidsg, line 834: Does this name need to change?

srh-sloan: No - they were different elements and not related but the diff is putting them next to each other I think coz of line numbers

gidsg (approved):
LGTM 👍

PR #136, opened 2022-09-20
FS-1501: secure cookies

gidsg:
Based on the IT Healthcheck we need to set secure properties on the session cookie.

### Screenshot after changes on dev
Screenshot 2022-09-20 at 12 42 27


srh-sloan (approved):
LGTM

PR #139, opened 2022-09-23
Changing order of applications on dashboard page from status to last edited

stevewalton28:
Tweaking order of applications on dashboard page
Tweaking font size of tag text


gidsg (approved):
LGTM 👍

PR #143, opened 2022-09-30
BAU: Add error handling to submit route

harryyo:
Adding error handling so an exception is raised on frontend is there is a failed submission response from application-store

### DIFF
--- a/app/default/routes.py
+++ b/app/default/routes.py
@@ -249,12 +249,25 @@ def submit_application():
         json=payload,
     )
     submitted = submission_response.json()
+    application_id = submitted.get("id")
+    application_reference = submitted.get("reference")
+    application_email = submitted.get("email")
+    if not application_reference or not application_email:
+        raise Exception(
+            "Unexpected response from application store when submitting application"
+            " application: "
+            + str(application_id) +
+            "application-store-response: "
+            + str(submission_response)
+        )
+
     response_weeks = 8
+
     return render_template(
         "application_submitted.html",
-        application_id=submitted.get("id"),
-        application_reference=submitted.get("reference"),
-        application_email=submitted.get("email"),
+        application_id=application_id,
+        application_reference=application_reference,
+        application_email=application_email,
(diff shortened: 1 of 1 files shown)


app/default/routes.py
```diff
@@ -249,12 +249,25 @@ def submit_application():
...
+    application_id = submitted.get("id")
+    application_reference = submitted.get("reference")
+    application_email = submitted.get("email")
+    if not application_reference or not application_email:
+        raise Exception(
+            "Unexpected response from application store when submitting application"
+            " application: "
+            + str(application_id) +
+            "application-store-response: "
+            + str(submission_response)
```
gidsg, line 261: Can we check the response code from the POST further up? e.g. like we do e.g. in https://github.com/communitiesuk/funding-service-design-frontend/blob/880ebe5e55f8bb5c07076477db71b383a8684f0e/app/default/routes.py#L127 rather than parsing the response body?

gidsg: Also, we could do hard lookup and fail earlier rather than `.get()` which may return `None` and cause us problems later on?

harryyo: No probs, updated

app/default/routes.py
```diff
@@ -249,17 +249,19 @@ def submit_application():
         json=payload,
     )
     submitted = submission_response.json()
-    application_id = submitted.get("id")
-    application_reference = submitted.get("reference")
-    application_email = submitted.get("email")
-    if not application_reference or not application_email:
+    if submission_response.status_code != 201 or not submitted.get(
```
gidsg, line 252: Have we checked the API definitely returns a `201` in the success case?

harryyo: Yes I checked that

harryyo: https://github.com/communitiesuk/funding-service-design-application-store/blob/main/openapi/api.yml#L180

gidsg (approved):
LGTM 👍

PR #145, opened 2022-09-30
BAU: remove unused env var

gidsg:


harryyo (approved):
Thanks Gideon, sorry should have removed earlier.

PR #147, opened 2022-10-04
FS-1601: Allow separate UAT environment

gidsg:
We want to allow separate UAT and Production environments in terraform. Even if there is no config changes with production it is still useful to separate them e.g. for Sentry.
Note: for now config is shared with production, so no separate config file is needed


adamdavies1 (approved):
LGTM

harryyo (approved):
Great

PR #163, opened 2022-10-12
Fs 1558 auth fixes

srh-sloan:
Adds a check to confirm the current logged in user is the owner of the application they are trying to act on, to the following endpoints:
/tasklist
/continue_application
/submit_application


app/default/routes.py
```diff
@@ -216,7 +223,12 @@ def continue_application(application_id):
         f"Url the form runner should return to '{return_url}'."
     )
 
+    current_user = g.account_id
     application = get_application_data(application_id, as_dict=True)
+    application_owner = application.account_id
+    if current_user != application_owner:
+        abort(401, f"User {current_user} attempted to continue_application for application {application_id}, owned by {application_owner}")
```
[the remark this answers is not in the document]
srh-sloan: We would need to get the application id into that decorator... I will have a look but definitely better to put it in one place if we can.

app/default/routes.py
```diff
@@ -239,16 +251,35 @@ def continue_application(application_id):
...
+def format_payload_and_submit_application(application_id):
     payload = {"application_id": application_id}
     submission_response = requests.post(
         Config.SUBMIT_APPLICATION_ENDPOINT.format(
             application_id=application_id
         ),
         json=payload,
     )
-    submitted = submission_response.json()
+    submitted = submit_application(application_id)
```
harryyo, line 282: I haven't quite understood this line - does the submit_application() route call format_payload_and_submit_application() which then calls submit_application() again?

tests/conftest.py
```diff
@@ -51,7 +51,9 @@ def app():
     @pytest.mark.uses_fixture('live_server')
     :return: An instance of the Flask app.
     """
-    yield create_app()
+    app = create_app()
+    # app.config["WTF_CSRF_ENABLED"] = False
```
harryyo, line 55: Can this comment go?

harryyo (commented):
Great, just two comments...

app/default/routes.py
srh-sloan: No, great spot, thanks. Line 282 isn't meant to be there - error on my part during the rebase.

app/default/routes.py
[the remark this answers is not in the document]
srh-sloan: Addressed now... the decorators are local to frontend atm, will move them to utils in a subsequent PR

app/default/routes.py
```diff
@@ -26,6 +29,32 @@
...
+        application_id = kwargs["application_id"]
+        return check_application_owner(application_id, f, *args, **kwargs)
+    return decorator
+
+def verify_application_owner_post_local(f):
+    @wraps(f)
+    def decorator(*args, **kwargs):
+        application_id = request.form["application_id"]
+        return check_application_owner(application_id, f, *args, **kwargs)
+    return decorator
```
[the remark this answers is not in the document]
srh-sloan: Yeah, good idea - done that now 👍

app/default/routes.py
```diff
@@ -26,6 +29,38 @@
 default_bp = Blueprint("routes", __name__, template_folder="templates")
 
 
+# TODO Move the following method into utils. Utils will need a way of accessing application data.
+
+def verify_application_owner_local(f):
+    """
+    This decorator determines whether the using trying to access an application is
```
Adam-W1, line 36: i think you meant user here

app/default/routes.py
```diff
@@ -301,6 +346,14 @@ def internal_server_error(error):
     current_app.logger.error(f"Encountered 500: {error}")
     return render_template("500.html"), 500
 
+@default_bp.errorhandler(401)
+def unauthorised_error(error):
+    current_app.logger.error(f"Encountered 401: {error}")
+    round_data = get_round_data_fail_gracefully(
```
Adam-W1, line 352: We use a mix of get_round_data and get round_data_gracefully, am I right in thinking we use the graceful get when we are due to send the info to the notification store - so the error surfaces there instead?

Adam-W1 (commented):
Looks great :) just a couple small comments

app/default/routes.py
srh-sloan: So the graceful one was specifically introduced for the error pages - so if it errors getting round data it just returns and empty object rather than throw an error. Then the error page deals with an empty error. Otherwise we had the situation where trying to display the error page could result in an error...

Adam-W1: ahh gotcha, okay thanks for clarifying 👍

Adam-W1 (approved):
Nice 👌

PR #167, opened 2022-10-14
fs-1625-adding-error-handling-for-csrf-expiry

gio-karunakaran:


app/default/routes.py
```diff
@@ -17,7 +17,8 @@
 from flask import request
 from flask import url_for
 from flask_wtf import FlaskForm
-from fsd_utils.authentication.decorators import login_required
+from fsd_utils.authentication.decorators import login_required, login_requested
```
harryyo, line 20: Has black run on this and allowed comma separated imports?

harryyo: 😮

harryyo (approved):
Great - only my shock horror that black has allowed comma separated imports! (I have nothing against them...!)

app/default/routes.py
gio-karunakaran: it did! I'm surprised too! let me separate them out for consistency's sake at least 👍🏾

PR #169, opened 2022-10-14
FS-1638 Increase form-runner disk-quota

harryyo:


manifest-form-runner.yml
```diff
@@ -4,6 +4,7 @@ applications:
   docker:
     image: ghcr.io/communitiesuk/digital-form-builder/runner:latest
   memory: 528M
+  disk_quota: 1024M
```
gidsg, line 7: Can we bump this to 2G instead? I think the image is close to 1GB already, so we may run into the same issue now or shortly.

harryyo: No probs - upped to 2G

gidsg (approved):

PR #170, opened 2022-10-17
FS-1638 Update form-runner-manifest env vars

harryyo:
Add SINGLE_REDIS = True to form runner test manifest


SentientHousePlant (approved):
nice work 💯

PR #173, opened 2022-10-20
Improved error handling on submit

srh-sloan:


app/default/routes.py
```diff
@@ -287,14 +288,15 @@ def submit_application():
...
-    current_app.logger.warning(f"Encountered 404: {error}")
+    current_app.logger.warning(f"Encountered 404 against url {request.path}: {error}")
     round_data = get_round_data_fail_gracefully(
         Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID
     )
     return render_template("404.html", round_data=round_data), 404
 
 
 @default_bp.errorhandler(500)
+@default_bp.errorhandler(Exception)
```
gidsg, line 299: What effect does this extra decorator have?

srh-sloan: It means if an exception is thrown rather than explicitly using `abort(500)`, this error handler still gets called

srh-sloan: So on line 268, we get the branded error page rather than a generic one (or the debug view in development)

gio-karunakaran (approved):
LGTM 👍🏾

gidsg (approved):

PR #179, opened 2022-10-21
Fs 1738 localise globals

donshellywork:
https://digital.dclg.gov.uk/jira/browse/FS-1738?workflowName=Funding+Services+%28FS%29&stepId=1

Added Welsh translation to the global elements (Header and Footer).
Added the translation files and file structure.
Added python logic to create_app.py that initializes the language support.
Added python logic that interprets the language via manual URL changes e.g. localhost:5000/language/cy
Correct Welsh translations are still needed, they have been requested via the excel spreadsheet


babel.cfg
```diff
@@ -0,0 +1,2 @@
+[python: app/**.py]
+[jinja2: **/templates/**.html]
```
gidsg, line 2: Do we need both `babel.cfg` and `config/babel.cfg` what is the difference? Can one be deleted?

gio-karunakaran: sorry they're the same file and the config one should stay, I'm removing the root babel file 👍🏾

donshellywork: @gidsg I didn't notice that, it was for tidiness. Shall we move to root?

gio-karunakaran: Moving it to the root for now as discussed, as so to be consistent with documentation: https://github.com/communitiesuk/funding-service-design-utils#translations

AbuNavsa (approved):
LGTM

gidsg (changes requested):
As discussed please use the jinja components rather than replacing with HTML.

app/templates/header.html
```diff
@@ -38,9 +36,18 @@
     </div>
 </header>
 
-
+<!-- Macro call removed for translation syntax to function correctly -->
 <div class="govuk-width-container">
-    {{ govukPhaseBanner({'tag': {'text': 'beta'}, 'html':
-    'This is a new service – your <a href="https://forms.office.com/Pages/ResponsePage.aspx?id=EGg0v32c3kOociSi7zmVqN48ORk8WN5LlJITE3Swt-lURUNCR0dHMjgxWFZOMTMxQzlOTVIxVkQ0Sy4u" class="govuk-link">feedback</a> will help us to improve it.'
-    })}}
```
donshellywork, line 45: @gidsg Ive tried to add gettext to this and babel does not capture the strings. Can you confirm that i should use the html instead of nunjuxks?

donshellywork: fixed, no need to go over this

gidsg (approved):
LGTM 👍 🏴󠁧󠁢󠁷󠁬󠁳󠁿

app/create_app.py
```diff
@@ -71,7 +76,7 @@ def inject_global_constants():
         return dict(
             stage="beta",
             service_title=(
-                "Apply for funding to save an asset in your community"
+                gettext(u'Apply for funding to save an asset in your community')
```
gidsg, line 79: Think this `u` prefix is redundant and can be safely removed.

gidsg (changes requested):
Looks like this is missing the compiled translation files? Please can you follow the steps here https://github.com/communitiesuk/funding-service-design-frontend/blob/welsh/docs/i18n.md and commit the generated files?

SentientHousePlant (dismissed):
Looks good but please address Gideons comment :)

app/create_app.py
SentientHousePlant: ☝️ I agree with this. The 'u' prefix means that a string should be interpreted as a unicode string, and 'b' means to represent the string as raw bytes in the backend - and python 3 uses unicode by default. 

Fun fact: This means that `b"😳"` would actually raise an error, since their is no bytes/ASCII representation of 😳.

donshellywork: @SentientHousePlant you're my favorite

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -27,37 +27,45 @@ msgstr "Gwneud cais am gyllid i achub ased yn eich cymuned"
...
 
 #: app/templates/header.html:46
 msgid "feedback"
-msgstr "adborth"
+msgstr "welsh needed"
 
 #: app/templates/header.html:46
 msgid " will help us to improve it."
-msgstr " yn ein helpu i'w wella."
+msgstr " welsh needed."
```
gidsg, line 70: Can we leave these blank rather than "Welsh needed"? It is better to show english if we don't have a translation rather than test data.

gio-karunakaran: have left untranslated data blank now 👍🏾

gidsg (approved):
Nice one, looks good to me 👍 🏴󠁧󠁢󠁷󠁬󠁳󠁿

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,70 @@
...
+
+#: app/templates/dashboard.html:10
+#, python-format
+msgid " %(count)s application "
+msgid_plural " %(count)s applications "
+msgstr[0] " %(count)s "
+msgstr[1] " %(count)s "
+msgstr[2] " %(count)s "
+msgstr[3] " %(count)s "
+msgstr[4] " %(count)s "
```
gidsg, line 34: We should have the translation for applications here, or leave blank.

gio-karunakaran: leaving blank for the time being as this is not part of this ticket 👍🏾

PR #203, opened 2022-11-01
fs-1831-creating-an-application-in-welsh

gio-karunakaran:


app/default/routes.py
```diff
@@ -158,6 +159,7 @@ def new():
             "account_id": account_id,
             "round_id": request.form["round_id"] or Config.DEFAULT_ROUND_ID,
             "fund_id": request.form["fund_id"] or Config.DEFAULT_FUND_ID,
+            "language": get_lang() or Config.DEFAULT_LANGUAGE,
```
gidsg, line 162: Don't think this will be necessary, `get_lang()` will always return something, my concern was more around fetching the the language of existing applications created prior to the migration.

gio-karunakaran: have removed the default lang envvar 👍🏾

adamdavies1 (commented):
Just a couple of things. Some straggling white spaces in there a well :)

app/create_app.py
```diff
@@ -11,11 +11,11 @@
 from flask_talisman import Talisman
 from flask_wtf.csrf import CSRFProtect
 from fsd_utils import init_sentry
-from fsd_utils.locale_selector.get_lang import get_lang
+from fsd_utils import LanguageSelector
```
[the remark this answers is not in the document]
gio-karunakaran: yep that's used to initialize the language selector on the flask app (line 33) 😄

app/default/routes.py
```diff
@@ -1,5 +1,6 @@
 from functools import wraps
 from http.client import METHOD_NOT_ALLOWED
+
```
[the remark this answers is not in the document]
gio-karunakaran: have removed! 👍🏾

app/models/application.py
```diff
@@ -44,25 +40,47 @@ def from_dict(cls, d: dict):
             }
         )
 
-    def create_sections(self, application):
+    @staticmethod
+    def create_blank_sections(fund_id, round_id, language):
```
gidsg, line 44: Could we add a test for this function?

app/models/application.py
```diff
@@ -44,25 +40,47 @@ def from_dict(cls, d: dict):
...
+        sections = []
```
gidsg, line 45: Could this be removed, as seems to be created further down (in line 56?)

app/models/application.py
gidsg, line 44: Could the config be passed in with a defaulted argument to make this more testable?

app/models/application.py
```diff
@@ -44,25 +40,47 @@ def from_dict(cls, d: dict):
...
+        try:
+            sections_config = FORMS_CONFIG_FOR_FUND_ROUND[
+                ":".join([fund_id, round_id])
+            ]
+        except IndexError:
+            current_app.logger.error(
+                f"FORM CONFIG for FUND:{fund_id} and ROUND:{round_id} does not"
+                " exist"
+            )
+        sections = [
```
gidsg, line 56: Could return this directly rather than assigning?

app/models/application.py
gio-karunakaran: removing this 👍🏾

tests/test_tasklist.py
```diff
@@ -76,3 +76,7 @@ def test_tasklist_for_submit_application_route(
     assert b"Application complete" in response.data
     assert b"Return to your applications page" in response.data
     assert b"test@example.com" in response.data
+
+
+def test_get_blank_sections():
+
```
gidsg: This is just an empty function, are we planning on adding a test here?

ramsharma-prog: Sorry forgot to remove this test. No need for test in here as we moved the static function to class function & that class is being tested.

app/models/application.py
gio-karunakaran: we have now changed this to a normal instance function and the Class is already being tested so this is being covered 👍🏾 :)

tests/test_tasklist.py
gio-karunakaran: I think this is an old push that didn't get refreshed by Github? 🤔

gidsg (approved):
LGTM 👍

PR #204, opened 2022-11-01
Fs 1798 unique field ids

Adam-W1:
Test to meet:
As an assessor I need to access my assessments, so they need to be organized clearly. To organize these clearly, these to uniquely specify them and devs need to be able to map them. To achieve this, we need to guarantee the keys are unique.

https://digital.dclg.gov.uk/jira/browse/FS-1798


tests/test_json_field_ids.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+    return form_json_with_field_ids
+
+
+def are_strings_unique(string_one, string_two):
+    if string_one == string_two:
+        return False
+    else:
+        return True
```
gidsg, line 33: Could we simplify this as follows?
```suggestion
    return string_one != string_two
```

gidsg:
What did we decide to do about Welsh versions? Should we allow them to have the same keys, so that they could be used in the Assessment service in future?

tests/test_json_field_ids.py
Adam-W1: yes - good shout 💯

Adam-W1:
> What did we decide to do about Welsh versions? Should we allow them to have the same keys, so that they could be used in the Assessment service in future?

Thats the most recent thinking, so we could place welsh forms into their own directory and test accordingly

gidsg:
> Thats the most recent thinking, so we could place welsh forms into their own directory and test accordingly

I think that would work for the test, but does the form runner & docker file expect the forms in a certain location? If so we might need to tweak other things to make that work?

gidsg:
@Adam-W1 don't see anything in here specific to Welsh forms, so will they not erroneously be flagged as duplicates? I see some have been moved to https://github.com/communitiesuk/funding-service-design-frontend/tree/main/form_jsons/public/welsh_forms now, do we need to run a version of the test against that folder. Or maybe move the english ones into `en` and welsh ones into `cy` folders?

Adam-W1:
> @Adam-W1 don't see anything in here specific to Welsh forms, so will they not erroneously be flagged as duplicates? I see some have been moved to https://github.com/communitiesuk/funding-service-design-frontend/tree/main/form_jsons/public/welsh_forms now, do we need to run a version of the test against that folder. Or maybe move the english ones into `en` and welsh ones into `cy` folders?

yes exactly that @gidsg. I spoke with @ramsharma-prog regarding this, you are correct we need a 'cy' folder and an 'en' folder, didnt realise this was now on main so ill rebase and make the en folder on this branch. Then we just need to point the test function at each folder independently. Ill update 👍

tests/test_json_field_ids.py
```diff
@@ -74,13 +75,26 @@ def check_for_duplicate_field_ids_across_form_jsons():
 
 
 def test_check_form_json_for_duplicate_field_ids():
```
gidsg, line 77: Rather than hardcoding and duplicating the english and welsh variants here, could we use parameterisation e.g. https://docs.pytest.org/en/6.2.x/parametrize.html and pass through `en` and `cy` parameters?

Adam-W1: A neat feature of pytest, thank you 💯

gidsg (approved):
LGTM 🏁 👍

PR #211, opened 2022-11-09
Fs 1737 localise applicant dashboard

gio-karunakaran:


AbuNavsa (approved):
LGTM

app/templates/dashboard.html
```diff
@@ -24,58 +24,57 @@ <h1 class="govuk-heading-xl">Your applications</h1>
...
         {% endif %}
 
+        {% set row = ns.rows.append(
+        [
+            {
+                'text': application.project_name if application.project_name else gettext('Untitled project')
+            },
+            {
+                'text':  govukTag({
+                    'text': application.status|snake_case_to_human ,
```
tferns, line 37:
```suggestion
                    'text': application.status.name|snake_case_to_human ,
```

This is an Enum so need to do `.name` or the UI is throwing an error for me

tferns: Also - should we find a way to translate the statuses? Seems since they're based on the Enum, and being formatted using `snake_case_to_human` they will always be English

app/templates/partials/start-application-button.html
```diff
@@ -1,14 +1,15 @@
...
-    {{ govukButton({
-    'text': "Start new application",
-    'classes': "govuk-button--secondary" if applications|length>0
-    }) }}
-</form>
+    <form method="POST" action={{url}} >
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+        <input type="hidden" name="account_id" value="{{ account_id }}">
+        <input type="hidden" name="fund_id" value="{{ fund_id }}">
+        <input type="hidden" name="round_id" value="{{ round_id }}">
```
tferns, line 8: This macro doesn't (and didn't, so not specifically related to your PR) have access to these variables afaik

app/templates/partials/start-application-button.html
```diff
@@ -1,14 +1,15 @@
 {%- from "govuk_frontend_jinja/components/button/macro.html" import govukButton -%}
 
 {% macro startApplicationButton(form, url, applications) %}
```
tferns, line 3: Do we need to pass in the `account_id, fund_id & round_id` here?

tferns (changes requested):
Changes look good, requesting changes for the bug with the Enum, which is preventing the UI from loading.  The other comments seem like issues worth fixing, but not directly related to your PR, so maybe there are future stories to cover them?

app/templates/dashboard.html
tferns: Disregard this, it's actually a string.  I had mocked an Enum without realizing it wasn't being used

tferns (approved):
Approving but with comments on macro (unrelated to this PR)

app/templates/dashboard.html
```diff
@@ -5,9 +5,9 @@
 
 {% extends "base.html" %}
 {% block content %}
-    <h1 class="govuk-heading-xl">Your applications</h1>
+    <h1 class="govuk-heading-xl">{% trans %}Your applications{% endtrans %}</h1>
     <p class="govuk-body">
-        You have started {% trans count=applications|length %} {{ count }} application {% pluralize %} {{ count }} applications {% endtrans %} using this email address.
+        {% trans %}You have started{% endtrans %}&nbsp;{% trans count=applications|length %}{{ count }} application{% pluralize %}{{ count }} applications{% endtrans %}&nbsp;{% trans %}using this email address{% endtrans %}.
```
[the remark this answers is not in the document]
gio-karunakaran: yes that's right! these were put here deliberately and we should probably start using more consistently? 🤔

app/templates/partials/start-application-button.html
gio-karunakaran: no need to pass these variables 👍🏾

app/templates/partials/start-application-button.html
gio-karunakaran: no need to pass these variables 👍🏾

PR #212, opened 2022-11-10
BAU: fix fund casing

gidsg:
Previously we were erroneously applying a filter on the already human readable fund name, which turned it to lower-case

## Before 
Screenshot 2022-11-10 at 10 07 46

## After
Screenshot 2022-11-10 at 10 09 26

### DIFF
--- a/app/templates/tasklist.html
+++ b/app/templates/tasklist.html
@@ -11,7 +11,7 @@
 
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-            <span class="govuk-caption-l">{{ application_meta_data["fund_name"]}} {{ application_meta_data["round_name"] }}</span>
+            <span class="govuk-caption-l">{{ application_meta_data["fund_name"]}} {{ application_meta_data["round_name"]|lower }}</span>
             <h1 class="govuk-heading-xl">Application for funding to save an asset in your community</h1>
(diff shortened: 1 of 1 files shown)


app/templates/tasklist.html
```diff
@@ -11,7 +11,7 @@
 
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-            <span class="govuk-caption-l">{{ application_meta_data["fund_name"]|kebab_case_to_human }} {{ application_meta_data["round_name"] }}</span>
+            <span class="govuk-caption-l">{{ application_meta_data["fund_name"]}} {{ application_meta_data["round_name"] }}</span>
```
Adam-W1, line 14: good spot! super picky but content brought this up last time I worked on casing, only the fund name is capitalized, not the word round as [here](https://fsd-funding-prototype-v2.herokuapp.com/cof-w2-v2/app-task-list-new#:~:text=Community%20Ownership%20Fund%20round%202)

gidsg: Should we change this upstream in the fund store as the casing comes from there?

gidsg: @Adam-W1 OK have added lower case filter now

Adam-W1 (approved):
👌

PR #215, opened 2022-11-15
FS-1759-localise-tell-us-your-email-page

ramsharma-prog:
Following translations are missing:

- Do not send any applications or attachments by email. We'll only accept applications through this service.


gio-karunakaran (approved):
LGTM 👍🏾

PR #216, opened 2022-11-15
remove-welsh-switcher-from-dashboard

ramsharma-prog:


Adam-W1 (approved):

gidsg (changes requested):
We should comment out the code rather than removing it, comments to follow

app/templates/dashboard.html
```diff
@@ -77,7 +77,7 @@ <h1 class="govuk-heading-xl">{% trans %}Your applications{% endtrans %}</h1>
 {% endif %}
 
 
-<p class="govuk-body">{% trans url=url_for('select_language', locale=('cy' if get_lang() == 'en' else 'en')) %} This service is also available <a href="{{ url }}" class="govuk-link">in Welsh (Cymraeg){% endtrans %}</a>.</p>
+<!-- TODO Add Welsh Switcher here -->
```
gidsg, line 80: Better if we comment out the two blocks in Jinja rather than removing.
Comments should be in Jinja rather than HTML e.g.

```
{# TODO: restore this once Welsh translation is completed #}
{#<p class="govuk-body">{% trans url=url_for('select_language', locale=('cy' if get_lang() == 'en' else 'en')) %} This service is also available <a href="{{ url }}" class="govuk-link">in Welsh (Cymraeg){% endtrans %}</a>.</p>#}
{##}
{#<div class="govuk-warning-text">#}
{#    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>#}
{#    <strong class="govuk-warning-text__text">#}
{#      <span class="govuk-warning-text__assistive">Warning</span>#}
{#      {% trans %}You cannot change the language of an existing application.{% endtrans %}#}
{#    </strong>#}
{#  </div>#}

ramsharma-prog: Yea, that's better. Done!

gidsg (approved):

PR #220, opened 2022-11-18
FS-1757: Add language availability notice

tferns:
Extracted a macro for language availability component, add to dashboard & landing page.

Relevant prototypes:
https://fsd-funding-prototype-v2.herokuapp.com/s22-application/applicant-dashboard
https://fsd-funding-prototype-v2.herokuapp.com/s20-application/invite-page

Images:
https://user-images.githubusercontent.com/117724519/202692956-d0878580-5f09-4c03-933b-15ebbf8987f0.png
https://user-images.githubusercontent.com/117724519/202693022-7316718f-a7ab-4154-841c-24b75b969a86.png

### DIFF
--- a/app/templates/dashboard.html
+++ b/app/templates/dashboard.html
@@ -2,6 +2,7 @@
 {%- from "govuk_frontend_jinja/components/table/macro.html" import govukTable -%}
 {%- from "govuk_frontend_jinja/components/button/macro.html" import govukButton -%}
 {% from "partials/start-application-button.html" import startApplicationButton %}
+{% from "partials/available_in_language.html" import available_in_language %}
 
 {% extends "base.html" %}
 {% block content %}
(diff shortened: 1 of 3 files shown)


app/templates/partials/available_in_language.html
```diff
@@ -0,0 +1,13 @@
+{% macro available_in_language(inset, warn) %}
```
tferns, line 1: With boolean traps, in python I'd force keyword arguments by swallowing positional arguments:

```python
def available_in_language(*, inset, warn):
    ...
````

Which would throw an error unless you explicitly specify keywords:
```
available_in_language(True, True)
  TypeError: available_in_language() takes 0 positional arguments but 2 were given
```

This doesn't seem to be available for jinja macros however :(

File: app/templates/dashboard.html [hunk 2]
```diff
@@ -79,16 +80,7 @@ <h1 class="govuk-heading-xl">{% trans %}Your applications{% endtrans %}</h1>
...
-{##}
-{#<div class="govuk-warning-text">#}
-{#    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>#}
-{#    <strong class="govuk-warning-text__text">#}
-{#      <span class="govuk-warning-text__assistive">Warning</span>#}
-{#      {% trans %}You cannot change the language of an existing application.{% endtrans %}#}
-{#    </strong>#}
-{#  </div>#}
-
+{{ available_in_language(inset=True, warn=True )}}
```
gidsg, line 83: Rather than not merging can we comment this out and add a a TODO to restore once translation is complete? (similar to deleted block)

(This way we could get it merged and avoid possible bit-rot)

tferns: Sure.  Good point

tferns: I've done this in https://github.com/communitiesuk/funding-service-design-frontend/pull/220/commits/a3910f0040c554ea0b5683544f767282d15cd264

tferns:
> LGTM 👍

Thanks for the review, I have force pushed a small change to remove the inset from the dashboard in https://github.com/communitiesuk/funding-service-design-frontend/commit/11ce2290b910c3b0b905291d2513ab9c78796b65

tferns:
Added a patch file to apply uncomment the changes when we need to: https://digital.dclg.gov.uk/jira/secure/attachment/112524/112524_patch.patch

PR #227, opened 2022-11-29
FS-1981: Fix translations for homepage

tferns:
**Summary:** 

Thought to have been fixed in https://github.com/communitiesuk/funding-service-design-frontend/pull/223, however branch wasn't updated with `main`, a small indentation change had happened in the html file, because the translation tag was indented it had another tab, and wasn't being matched in the `.po` anymore.

Going forward we should probably avoid newlines or tabs in translation tags.

- Add script to easily remove newlines & tabs from translation tags
- Add pre-determined invoke commands so no need to specify arguments each time
- Update .pot .po & .po to reflect translation changes to index.html
- Update requirements to include invoke


rayhan15243 (approved):

messages.pot
```diff
@@ -108,58 +108,50 @@ msgstr ""
 msgid " will help us to improve it."
 msgstr ""
 
-#: app/templates/index.html:8
-msgid "Community Ownership Fund "
-msgstr ""
-
-#: app/templates/index.html:11
-msgid "Community Ownership Fund"
-msgstr ""
```
tferns, line 117: These are also no longer used

messages.pot
```diff
@@ -108,58 +108,50 @@ msgstr ""
...
 msgid "Get help "
 msgstr ""
 
-#: app/templates/index.html:35
+#: app/templates/index.html:33
 msgid "If you need help, email us at"
 msgstr ""
 
-#: app/templates/index.html:39
+#: app/templates/index.html:37 app/templates/tasklist.html:74
```
tferns, line 151: When we remove newlines and tabs, files re-use the same translation and stack

gidsg (approved):
LGTM 👍

gidsg:
Should we replicate these tasks in the authenticator repo as it also has translations?

tferns:
> Should we replicate these tasks in the authenticator repo as it also has translations?

Yeah I think that sounds like a good idea.  Will raise a PR for that 👍

Maybe for common scripts, we could start adding to fsd-utils or a unique project specifically for CLI tools? fsd-cli-tools?  That way we would only have to maintain them in once place and each project could install and utilize them, could also export some pre-commit hooks.

gidsg:
@tferns yes, potentially some of this could be in utils, would have to make sure it really is generic though, as e.g. authenticator stores its translations in a different folder I believe.

PR #232, opened 2022-12-08
FS-2046: Pull cookie domain from env var

tferns:
- Change default config, pull cookie domain from COOKIE_DOMAIN env var
- If COOKIE_DOMAIN env var is not set, it will default to None
- Update dev & test manifest to include COOKIE_DOMAIN

This is needed to persist the `language` cookie when going from frontend -> authenticator.


gidsg (approved):
LGTM 👍

PR #237, opened 2023-01-16
FS-2172: use shared workflow for tag

gidsg:
Use shared workflow for tagging, instead of copy-pasting between repos.

Nature of github actions makes this very tricky to test on a branch, so plan is to merge this in, test and then revert of fix-forward.


stevewalton28 (approved):

Adam-W1 (approved):
nice

PR #240, opened 2023-01-18
BAU Refactor routes

srh-sloan:
During initial investigation of FS-2183 (now moved to a later sprint) the `routes.py` seemed very big and hard to navigate. Have split into smaller routes files to make navigation and future work easier.


Adam-W1 (approved):
Love a good tidy up 🧹 ✨

app/default/application_routes.py
```diff
@@ -0,0 +1,232 @@
...
+from flask import request
+from flask import url_for
+from flask_wtf import FlaskForm
+from fsd_utils.authentication.decorators import login_required
+from flask_babel import force_locale
+from app.default.routes import current_datetime_after_given, current_datetime_before_given
+
+application_bp = Blueprint("application_routes", __name__, template_folder="templates")
+
+def verify_application_owner_local(f):
```
gidsg, line 30: should this sit outside of routes?

srh-sloan: Yes there's a TODO I missed when copying for it to move to utils with a couple of other methods. Will do those in a separate PR.

PR #242, opened 2023-01-20
Updating forms, styles not needed

adamdavies1:
A recent push to the form runner has made the workaround for making the hints look like labels redundant
https://github.com/communitiesuk/digital-form-builder/commit/92f715a8e5aba6bb5c6433f66a892667dcc5ac3a


gidsg (approved):
Looks good, do we need to bump the version if only `form_jsons` are changing?

adamdavies1:
> Looks good, do we need to bump the version if only `form_jsons` are changing?

Im not actually sure. I think we might but ill see if the merge has any affect

PR #243, opened 2023-01-23
FS-2190: unique title tags

gidsg:
As highlighted in the accessibility audit, pages should have unique titles, previously most pages just used the `service_title`

This adds a `pageHeading` variable in the base template which is used in the <title> tag if set and as the `h1` to avoid duplication and encourage re-use of this pattern in future.

### DIFF
--- a/app/templates/tasklist.html
+++ b/app/templates/tasklist.html
@@ -6,7 +6,7 @@
 {% block beforeContent %}
     <a href="{{ url_for('account_routes.dashboard') }}" class="govuk-back-link">Back to your applications</a>
 {% endblock %}
-{% block pageTitle %}{% trans %}Task List{% endtrans %} - {{ service_title }}{% endblock %}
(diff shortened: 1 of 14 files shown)


app/templates/tasklist.html
```diff
@@ -6,7 +6,7 @@
 {% block beforeContent %}
     <a href="{{ url_for('account_routes.dashboard') }}" class="govuk-back-link">Back to your applications</a>
 {% endblock %}
-{% block pageTitle %}{% trans %}Task List{% endtrans %} - {{ service_title }}{% endblock %}
+{% set pageHeading %}{% trans %}Task List{% endtrans %}{% endset %}
```
Adam-W1, line 9: should this be used on page?

gidsg: There is an inconsistency in this case, Task List was set as the title which is different than the `H1` so I went with Task List as that's more specific.

Adam-W1 (approved):
looks good 👍

PR #250, opened 2023-01-26
Bau deploy order

srh-sloan:
Updates the workflow to run unit tests before deploy, then just the accessibility tests after deploy


gidsg (approved):
👍

PR #259, opened 2023-02-09
BAU: All questions round 2 window 3 page

srh-sloan:
Uses endpoint added in https://github.com/communitiesuk/funding-service-design-fund-store/pull/50

Updates:
- Change of URL to get to page showing all questions (now /all_questions/<fund_short_name>/<round_short_name> eg. `/all_questions/cof/r2w2`
- Redirect added so old url works (`/cof_r2w2_all_questions`)
- Goes along with changes to the fund store to be able to retrieve by short_name, and with the authenticator so it dynamically generates the link to the all questions page

### DIFF
--- a/app/default/data.py
+++ b/app/default/data.py
@@ -121,6 +121,23 @@ def get_round_data(fund_id, round_id, as_dict=False):
         return round_response
 
 
+def get_round_data_by_short_names(fund_short_name, round_short_name):
+    request_url = Config.GET_ALL_ROUNDS_FOR_FUND_ENDPOINT.format(
+        fund_id=fund_short_name
+    )
+    response = get_data(
+        request_url, {"language": get_lang(), "use_short_name": "true"}
+    )
+    return next(
+        (
+            round
+            for round in response
+            if str.upper(round["short_name"]) == str.upper(round_short_name)
+        ),
+        None,
+    )
+
+
 def get_round_data_fail_gracefully(fund_id, round_id):
     try:
         language = {"language": get_lang()}
--- a/app/default/data.py
+++ b/app/default/data.py
@@ -132,7 +132,7 @@ def get_round_data_by_short_names(fund_short_name, round_short_name):
         (
             round
             for round in response
-            if str.upper(round["short_name"]) == str.upper(round_short_name)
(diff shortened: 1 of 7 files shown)


app/default/data.py
```diff
@@ -121,6 +121,23 @@ def get_round_data(fund_id, round_id, as_dict=False):
...
+        fund_id=fund_short_name
+    )
+    response = get_data(
+        request_url, {"language": get_lang(), "use_short_name": "true"}
+    )
+    return next(
+        (
+            round
+            for round in response
+            if str.upper(round["short_name"]) == str.upper(round_short_name)
```
tferns, line 135: There are some nuances with characters that can cause bugs if doing case comparison with `upper` or `lower` - so the standard way to do caseless matching is [casefold](https://docs.python.org/3/library/stdtypes.html#str.casefold).  I don't know if there will be any weird characters in the rounds `short_name` that will cause an issue, but it's good to use `casefold` as a standard anyways imo.

```suggestion
            if str.casefold(round["short_name"]) == str.casefold(round_short_name)
```

app/default/data.py
```diff
@@ -121,6 +121,23 @@ def get_round_data(fund_id, round_id, as_dict=False):
...
+        ),
+        None,
```
tferns, line 137: Ah, is `None` a default? That's pretty cool, however it might just defer to a `KeyError` exception:

https://github.com/communitiesuk/funding-service-design-frontend/blob/a572e6c033fa953b0b8acfeda57c69c3cdfc47e3/app/default/content_routes.py#L29

```bash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'title'
```

Don't know if this will ever happen though, so not sure how valuable it is to make the error more verbose... I think with the microservices architecture we're using there are quite a lot of scenarios like this due to the high amount of requests between services.  Thought I'd mention anyways since it caught my eye, but not a big deal.

tferns (approved):
Looks good but 1 suggestion and 1 minor concern

app/default/data.py
srh-sloan: Thanks - added a check for round being None

PR #265, opened 2023-02-14
Updates to banner for r2w3

srh-sloan:
Updates to the window closed banner:
- Hide banner for R2W2 if they didn't apply in window 2
- Remove 'we will announce the next round shortly' text from banner
As per design requirements from Emma

### DIFF
--- a/app/default/account_routes.py
+++ b/app/default/account_routes.py
@@ -56,6 +56,8 @@ def build_application_data_for_display(applications: list[ApplicationSummary]):
             ]
             if not_yet_open:
                 continue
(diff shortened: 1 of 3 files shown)


app/default/account_routes.py
```diff
@@ -56,6 +56,8 @@ def build_application_data_for_display(applications: list[ApplicationSummary]):
             ]
             if not_yet_open:
                 continue
+            if past_submission_deadline and len(apps_in_this_round) == 0:
+                continue
```
gidsg, line 60: can we combine these conditions?

PR #270, opened 2023-02-15
BAU: Pass in Assessment frontend URL env var for e2e assessment

rayhan15243:
Added env var for assessment as it was causing the e2e assessment test to fail when the workflow was running via application frontend.
https://github.com/communitiesuk/funding-service-design-frontend/actions/runs/4182482703/jobs/7245810232


tferns (approved):
👍🚢

PR #272, opened 2023-02-16
Adding language column

adamdavies1:
Language column is now part of the application row. The language is at first in its short code version. It is the converted to its full name via the langcodes library 

image


app/templates/partials/applications_table.html
```diff
@@ -69,6 +72,9 @@
                 {
                     'text': gettext('Application Reference')
                 },
+                {
+                    'text': gettext('Language')
```
tferns, line 76: Are there Welsh translations for this?

app/templates/partials/applications_table.html
```diff
@@ -47,6 +47,9 @@
         {
             'html': application.reference
         },
+        {
+            'html': application.language
```
tferns, line 51: Not really cause of your PR, but do you know why these are `html` as opposed to `text` - guess it doesn't change much just curious, feel free to ignore this.

tferns (approved):

adamdavies1:
This is blocked for now whilst design has a discussion about whether or not this provides value

app/templates/partials/applications_table.html
```diff
@@ -2,59 +2,114 @@
...
-        },
-        {
-            'html': application.reference
-        },
-        {
-            'html': application_link
-        }
-    ])%}
-    {% endfor %}
+		{% if showLanguageColumn %}
```
tferns, line 33: This popped into my head after work so commenting it before I forget on Monday lol.

A way to do optional dict inside array based on a condition, to remove the big if blocks:

```python
def test(condition):
    return [
        {a:1},   
        *([{b:2}] if condition else []),
        {c:3},
    ]

test(True)   # [{a:1}, {b:2}, {c:3}]
test(False)  # [{a:1}, {c:3}]

```

app/models/language.py
```diff
@@ -0,0 +1,13 @@
...
+    languages = {
+        "en" : gettext("English"),
+        "cy" : gettext("Welsh")
+    }        
+    try:
+        translated_value = languages[value]
+    except KeyError:            
+        translated_value = value
+
+    return translated_value
```
tferns, line 13:
```suggestion
    return languages.get(value, value)
```

app/models/statuses.py
```diff
@@ -0,0 +1,24 @@
...
+}
+    try:
+        translated_value = statuses[value]
+    except KeyError:
+        current_app.logger.info(
+        f"'{translated_value}' is not in statuses dictionary"
+    )
+        return snake_case_to_human(value)
+  
+    return translated_value
```
tferns, line 24: indentation seems wrong here, and no line at end of file, could you run pre-commit?

app/models/statuses.py
```diff
@@ -0,0 +1,24 @@
...
+    "FLAGGED": gettext("Flagged"),
+    "STOPPED": gettext("Stopped"),
+}
+    try:
+        translated_value = statuses[value]
+    except KeyError:
+        current_app.logger.info(
+        f"'{translated_value}' is not in statuses dictionary"
+    )
+        return snake_case_to_human(value)
```
tferns, line 22:
```suggestion
    return statuses.get(value, snake_case_to_human(value))
```

app/default/account_routes.py
```diff
@@ -109,6 +115,7 @@ def dashboard():
         "dashboard.html",
         account_id=account_id,
         display_data=display_data,
+        showLanguageColumn=showLanguageColumn
```
tferns, line 118: Could we be consistent and use snake_case?

app/templates/partials/applications_table.html
adamdavies1: Unfortunately i havent been able to get this to work in nunjuck (not sure if it does). Have left it as it is for now

app/models/statuses.py
adamdavies1: Class look completely different now and the indentation should be fixed

app/default/account_routes.py
```diff
@@ -101,6 +101,12 @@ def dashboard():
         for application in application_store_response
     ]
 
+    current_app.logger.info(
+        f"Applications :'{applications}'"
+    )
+
```
gidsg, line 107: Can we remove this logging I guess it was for debugging but doesn't really make sense to log all applications without context.

app/templates/partials/applications_table.html
```diff
@@ -2,59 +2,113 @@
...
-        },
-        {
-            'html': application.reference
-        },
-        {
-            'html': application_link
-        }
-    ])%}
-    {% endfor %}
+		{% if show_language_column %}
```
gidsg, line 33: As discussed can we try and do this without repeating the whole table?

tferns: @adamdavies1  How about a Jinja filter?

```python
from jinja2 import Environment

def conditional_list(items):
    return [val for val, hide in items if not hide]

env = Environment()
env.filters['conditional_list'] = conditional_list

# Usage example:
items = [(1, False,), (2, True,), (3, False,),]
template = env.from_string("Filtered items: {{ items|conditional_list }}")
rendered = template.render(items=items)
print(rendered)  # Output: Filtered items: [1, 3]
```

Although it might be a bit annoying to have to specify a tuple for every element - but the function could probably be enhanced to take that into account and default anything else to just returning.

app/models/statuses.py
```diff
@@ -0,0 +1,13 @@
+from flask_babel import gettext
+
+def get_text(value: str):
```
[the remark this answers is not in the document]
adamdavies1: Change this to get_formatted

app/default/account_routes.py
adamdavies1: Removed this

app/models/statuses.py
```diff
@@ -0,0 +1,13 @@
...
+
+    statuses = {    
+    "NOT_STARTED": gettext("Not Started"),
+    "IN_PROGRESS": gettext("In Progress"),
+    "COMPLETED": gettext("Assessment complete"),
+    "QA_COMPLETED": gettext("QA complete"),
+    "FLAGGED": gettext("Flagged"),
+    "STOPPED": gettext("Stopped"),
+}
+    return statuses.get(value, value.replace("_", " ").strip().title())
```
tferns, line 13: This file isn't formatted correctly, could you run pre-commit on it?

The reason no newline at end of file could be a problem:

```bash
ThomasFerns@bash ~
$ echo -n 'x = 1' > t.py
ThomasFerns@bash ~
$ cat t.py
x = 1ThomasFerns@bash ~

ThomasFerns@bash ~
$ echo -n 'x = 1\n' > t.py
ThomasFerns@bash ~
$ cat t.py
x = 1
ThomasFerns@bash ~
```

Bunch of other things can become erroneous as well, the flake8 rule is `W292`

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -19,14 +19,46 @@ msgstr ""
...
+msgid "English"
+msgstr "Saesneg"
+
+#: app/models/language.py:6
+msgid "Welsh"
+msgstr "Cymraeg"
+
+#: app/models/statuses.py:9
+msgid "Not Started"
+msgstr "Rydych wedi dechrau"
```
tferns, line 40: This translates to "You have started" via google translate - is this correct?

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -19,14 +19,46 @@ msgstr ""
...
+
+#: app/models/statuses.py:10
+msgid "In Progress"
+msgstr "In Progress Welsh"
```
tferns, line 44: Is this translation missing? If so could we remove the translation until we have one, so it just shows "In Progress"

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -19,14 +19,46 @@ msgstr ""
...
+
+#: app/models/statuses.py:11
+msgid "Assessment complete"
+msgstr "o'r meini prawf asesu"
```
tferns, line 48: Is this translation correct? If it is then should we have a capital at the beginning?

```suggestion
msgstr "O'r meini prawf asesu"
```

app/translations/cy/LC_MESSAGES/messages.po
```diff
@@ -2401,3 +2448,10 @@ msgstr "Dechrau cais newydd"
 #: app/templates/partials/tasklist_section.html:7
 msgid "of the assessment criteria"
 msgstr "o'r meini prawf asesu"
+
+#~ msgid "English"
+#~ msgstr "Saesneg"
+
+#~ msgid "Welsh"
+#~ msgstr "Cymraeg"
+
```
tferns, line 2457: Are these redundant comments?

tests/test_application_summary.py
```diff
@@ -112,16 +113,19 @@
...
-    assert len(applications) == 3
-    assert applications[0].started_at.__class__.__name__ == "datetime"
-    assert str(applications[0].started_at.tzinfo) == "Europe/London"
-    assert applications[1].last_edited is None
+        applications = [
+            ApplicationSummary.from_dict(application)
+            for application in application_list
+        ]
+        assert len(applications) == 3
+        assert applications[0].started_at.__class__.__name__ == "datetime"
```
tferns, line 124: I'd maybe use an instance check here instead of a string assertion, I don't think this is too big of a deal but in theory there _could_ be other classes from different packages named `datetime` - very unlikely but the instance check is just more standard practice.  

```suggestion
        assert assert isinstance(applications[0].started_at, datetime.datetime)
```

tferns (approved):
Looks good!  Good work on this cause I know the syntax is a pain... I've left a few minor suggestions but nothing major.

Might be good to remember to squash commits as there has been quite a few?

app/translations/cy/LC_MESSAGES/messages.po
adamdavies1: Dont trust google translate haha. But this is wrong. we are waiting for translations for these still

tferns: Ah cool, it looked a bit longer than "Not Started" which is why I checked, if we're still waiting could we leave them empty for now?

app/translations/cy/LC_MESSAGES/messages.po
adamdavies1: Updated this to one of the ones we have already

app/translations/cy/LC_MESSAGES/messages.po
adamdavies1: These are auto generated

PR #285, opened 2023-02-23
FS-2309-display-fund-name-based-on-selected-language

ramsharma-prog:
This PR resolves a technical issue related to the display of fund and round names. It fixes the problem of incorrect translation for these names when users select English or Welsh on the platform. The fix ensures that fund and round names are accurately displayed in the user's selected language. Additionally, the PR also addresses the translation issue for static page fund names.

### Change description
_A brief description of the pull request_ 

- [x] Unit tests and other appropriate tests added or updated
- [ ] README and other documentation has been updated / added (if needed)- NA
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### Screenshots of UI changes.

**Tasklist page**
Screenshot 2023-02-23 at 11 00 27

**Static Page**
Screenshot static 2023-02-23 at 11 00 37


app/default/data.py
```diff
@@ -99,21 +99,19 @@ def get_applications_for_account(account_id, as_dict=False):
         return application_response
 
 
-def get_fund_data(fund_id, as_dict=False):
+def get_fund_data(fund_id, language=None, as_dict=False):
     fund_request_url = Config.GET_FUND_DATA_ENDPOINT.format(fund_id=fund_id)
-    language = {"language": get_lang()}
```
gidsg, line 104: I think we need to keep this for other language switching scenarios, but we should only do this if no language is passed-in to the function.

ramsharma-prog: @gidsg I am passing the  language from application and the application will always have a language so do we need it?

ramsharma-prog: I see what you mean. Function updated @gidsg 
Thank you

app/default/routes.py
```diff
@@ -29,15 +36,15 @@ def index():
...
-        service_url=Config.ENTER_APPLICATION_URL,
-        fund_name=fund_name,
-        round_title=round_title,
-        submission_deadline=submission_deadline,
-        is_past_submission_deadline=current_datetime_after_given_iso_string(
-            submission_deadline
-        ),
-        contact_us_email_address=contact_us_email_address,
-    )
+    with force_locale(get_lang()):
```
gidsg, line 39: Don't think we need `force_locale` here as `get_lang()` should be the default behaviour anyway?

ramsharma-prog: Yes!! Sorry should have removed since get_lang() is set be default

gidsg (approved):
LGTM 👍

PR #293, opened 2023-03-07
FS-2414-add-remaining-translation-to-tasklist-page

ramsharma-prog:
Screenshot 2023-03-07 at 16 15 32


rayhan15243:
@ramsharma-prog 81 changes? Where did they come from?

PR #295, opened 2023-03-16
FS-2390: Fix compatible html elements

tferns:
### Change description

For lists in html, we should ensure if we have alternative elements they're encapsulated in a list item.  This change does that on two instances that were raised by [axe dev tools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd).  Some translations weren't working because they included % signs, I've fixed them as well while I was in the area.

### How to test

Checkout the branch, re-start your docker container to ensure the new compiled translations are picked up.  Then visit the all questions page for [cof r2w3](http://localhost:3008/all_questions/COF/R2W3?lang=cy), you can observe the fixed translations.  You can also run the axe analysis using the extension I linked above. 

### Screenshots of axe analysis

image

Note: This also included pre-commit-ci changes, which fixed issues throughout the entire codebase, hopefully we'll only have these once per repo.


srh-sloan (commented):
You probably already tried it but rather than split translations would it make more sense to use the escaped version of %? So `Can provide 50&percnt; of the total`. Not sure if it would work though...

tferns:
> You probably already tried it but rather than split translations would it make more sense to use the escaped version of %? So `Can provide 50&percnt; of the total`. Not sure if it would work though...

I thought about this but didn't try it! I'll give it a shot.

tferns:
@srh-sloan Genius! Thanks for that recommendation, have added in https://github.com/communitiesuk/funding-service-design-frontend/commit/6c87693a47d940f9dd2403df71342d1b22da9833

srh-sloan (approved):
Looks good, nice one!

Adam-W1 (approved):

PR #304, opened 2023-04-06
Fs 2514 multi fund dashboard

srh-sloan:
https://digital.dclg.gov.uk/jira/browse/FS-2514

### Change description
Updating the dashboard page (frontend/account) to provide a 'view all applications' option, as per this mural https://app.mural.co/t/mhclg2837/m/mhclg2837/1675177182644/578e9019ed4a6da033b5de2183289ca7298e5f12?sender=u94fce4b3d87aef5380039386.

Dashboard has been split into 2 html pages - `dashboard_single_fund.html` is the current dashbaord, `dashboard_all.html` is for all applications.

- [ ] Unit tests and other appropriate tests added or updated
- [ ] README and other documentation has been updated / added (if needed)
- [ ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
test the dashboard as normal, a view all applications link should display on the header next to the signout link. If you want to see this feature toggled off (as it would be in production for now) change the feature config value in default config to use the prod feature config from common config (can check in production config to see how this is done)


app/default/account_routes.py
```diff
@@ -79,35 +126,75 @@ def build_application_data_for_display(applications: list[ApplicationSummary]):
...
-    application_store_response = get_applications_for_account(
-        account_id=account_id, as_dict=False
-    )
+    fund_short_name = request.args.get("fund")
+    round_short_name = request.args.get("round")
+
+    if fund_short_name and round_short_name:
+        # search for applications for this account AND
+        # this fund if fund is supplied, else get all
+        # applications for this account
```
srh-sloan, line 148: Comment needs updating - duplicate of case below

config/envs/test.py
```diff
@@ -15,3 +15,12 @@ class TestConfig(DefaultConfig):
...
     ).decode()
+
+    # Redis Feature Toggle Configuration
+    REDIS_INSTANCE_NAME = "funding-service-magic-links-test"
+    REDIS_INSTANCE_URI = (
+        DefaultConfig.VCAP_SERVICES.get_service_credentials_value(
+            "redis", REDIS_INSTANCE_NAME, "uri"
+        )
+    )
+    TOGGLES_URL = REDIS_INSTANCE_URI + "/0"
```
srh-sloan, line 26: `TOGGLES_URL` is the same everywhere so can we move to default?

srh-sloan (commented):
A few small comments but looking good

config/envs/default.py
```diff
@@ -169,4 +174,19 @@ class DefaultConfig:
     DEFAULT_FUND_ID = CommonConfig.COF_FUND_ID
     DEFAULT_ROUND_ID = CommonConfig.get_default_round_id()
 
-    FORMS_CONFIG_FOR_FUND_ROUND = CommonConfig.FORMS_CONFIG_FOR_FUND_ROUND
+    FORMS_CONFIG_FOR_FUND_ROUND = {
+        "47aef2f5-3fcb-abab-acb5-f0152b5f03c4:c603d114-5364-2222-a0c4-c41cbf4d3bbd": CommonConfig.COF_R2_ORDERED_FORMS_CONFIG,  # noqa
```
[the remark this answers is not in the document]
gio-karunakaran: sorry, great spot!

config/envs/test.py
gio-karunakaran: done!

tests/test_dashboard.py
```diff
@@ -0,0 +1,426 @@
+import json
```
tferns, line 1: @srh-sloan these tests are amazing 🔥 👍

tferns (approved):
Few comments! But looks good 🚢👍

app/default/account_routes.py
```diff
@@ -19,52 +22,96 @@
...
+        funds_to_show = [
+            fund
+            for fund in all_funds
+            if fund["short_name"].casefold()
+            == visible_fund_short_name.casefold()
+        ]
+    else:
+        funds_to_show = all_funds
+
+    if not funds_to_show or len(funds_to_show) == 0:
```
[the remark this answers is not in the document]
gio-karunakaran: good spot!

app/default/data.py
```diff
@@ -192,8 +204,9 @@ def get_round_data(fund_id, round_id, language=None, as_dict=False):
 
 def get_round_data_by_short_names(
     fund_short_name, round_short_name, as_dict=False
-):
+) -> Round:
```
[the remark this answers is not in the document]
gio-karunakaran: 👌🏾

config/envs/test.py
```diff
@@ -15,3 +15,11 @@ class TestConfig(DefaultConfig):
...
         environ.get("RSA256_PUBLIC_KEY_BASE64")
     ).decode()
+
+    # Redis Feature Toggle Configuration
+    REDIS_INSTANCE_NAME = "funding-service-magic-links-test"
+    REDIS_INSTANCE_URI = (
+        DefaultConfig.VCAP_SERVICES.get_service_credentials_value(
+            "redis", REDIS_INSTANCE_NAME, "uri"
+        )
+    )
```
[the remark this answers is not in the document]
gio-karunakaran: added!

app/default/account_routes.py
```diff
@@ -19,52 +22,96 @@
...
                     "is_past_submission_deadline": round_status.past_submission_deadline,  # noqa
                     "is_not_yet_open": round_status.not_yet_open,
                     "round_details": round,
-                    "applications": apps_in_this_round,
+                    "applications": apps_for_display,
                 }
             )
-            count_of_applications_for_visible_rounds += len(apps_in_this_round)
+            count_of_applications_for_visible_rounds += len(apps_for_display)
+        if len(fund_data_for_display["rounds"]) == 0:
```
Adam-W1, line 113: In Python, empty collections such as lists, dictionaries, and sets are considered "falsy".  So you can directly use "if not fund_data_for_display["rounds"]:" without needing to call the "len()" function.

gio-karunakaran: done!

tferns: +1 ^

```suggestion
        if not fund_data_for_display["rounds"]:
```

gio-karunakaran: sorted :D

tferns (approved):

app/default/account_routes.py
```diff
@@ -19,52 +22,96 @@
...
+def get_visible_funds(visible_fund_short_name):
+    """
+    Returns a list of funds matching the supplied short name
 
+    :param visible_fund_short_name: short name to look for
+    """
+    all_funds = get_all_funds()
+
+    if visible_fund_short_name:
+        funds_to_show = [
```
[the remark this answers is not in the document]
gio-karunakaran: one to check with @srh-sloan when she's back?

app/templates/partials/applications_table.html
```diff
@@ -27,7 +27,7 @@
 		{% endset %}
 
 		{% if not application.status == 'SUBMITTED' and not is_past_submission_deadline %}
-			{% set application_link = '<a class="govuk-link" href="{}">{}</a>'.format(url_for('application_routes.tasklist', application_id=application.id), gettext('Continue application')) %}
+			{% set application_link = '<a class="govuk-link" href="{}">{}</a>'.format(url_for('application_routes.tasklist', application_id=application.id, fund=fund_short_name, round=round_short_name), gettext('Continue application')) %}
```
Adam-W1, line 30: why do we pass fund and round here?

gio-karunakaran: so we can re-direct user back to the fund round page (/accont?fund=blah&round=blah) as opposed to the all applications page (/account) + we want to hide access to this page until we run concurrent live funds 👍🏾

app/templates/dashboard_all.html
```diff
@@ -0,0 +1,48 @@
...
+                        {{ fund["fund_data"]["name"] }}
+                    </span>
+                </h2>
+            </div>
+            <div id="funds-content-{{fund['fund_data']['short_name']}}" class="govuk-accordion__section-content" aria-labelledby="funds-heading-{{fund['fund_data']['short_name']}}">
+
+                {% for round in fund["rounds"] %}
+                    {% if not round["is_not_yet_open"] and not round["is_past_submission_deadline"] %}
+                        <span class="govuk-caption-m">{{ round["round_details"]["title"] }}</span>
+                        {% if round["is_past_submission_deadline"] %}
```
Adam-W1, line 36: I think this code in unreachable due to the first if statement. 
the only way to enter the if is if is_past_submission_deadline == False, so`and not False` on line 34 will get into the first block 
then `if False` on line 36 will prevent the second block ever executing

app/templates/dashboard_single_fund.html
```diff
@@ -31,7 +31,7 @@ <h2 class="govuk-heading-l">
                 {% endif %}
                     {{ applications_table(round["applications"], round["is_past_submission_deadline"], show_language_column) }}
                 {% if not round["is_past_submission_deadline"] %}
-                    {{ startApplicationButton(form, url_for('account_routes.new'), fund_id=fund["fund_id"], round_id=round["round_details"]["id"], existing_applications_count=round["applications"]|count) }}
+                    {{ startApplicationButton(form, url_for('account_routes.new'), fund_id=fund["fund_data"]["id"], round_id=round["round_details"]["id"], existing_applications_count=round["applications"]|count) }}
                 {% endif %}
             {% endif %}
         {% endfor %}
```
Adam-W1, line 37: I had the check this with Alan but...
_In general, it is not recommended to include complex logic or access complex data structures such as dictionaries or arrays directly within templates in most templating engines. This is because templates are meant to be focused on rendering and presentation logic, while the business logic and data manipulation should be handled outside of the template.

Instead, it is better to provide the template with the variables or data it needs from the external context, usually through a separate controller or view function. In this case, it would be preferable to provide the template with the variable fund_id directly from the controller or view function, rather than accessing fund["fund_data"]["id"] directly within the template._

app/create_app.py
```diff
@@ -102,6 +113,12 @@ def inject_global_constants():
             service_meta_author=(
                 "Department for Levelling up Housing and Communities"
             ),
+            toggle_dict={
```
Adam-W1, line 116: nice use of global constant with the toggles 👌

app/templates/dashboard_all.html
gio-karunakaran: you're right! have removed this now 👍🏾

app/templates/partials/applications_table.html
Adam-W1: ah okay so the logic doesnt exist on tasklist.html to pick this up yet?

app/templates/dashboard_single_fund.html
gio-karunakaran: Ah! although I do agree with Alan, this is the simplest (and i think only) way for us to loop through each round and use the appropriate/ dynamic IDs for the urls. We could generate all our info needed in the function calling the template but this will still have to be stored in a complex data structure for us to dynamically call the right IDs... 😢

app/default/account_routes.py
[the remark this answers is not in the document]
srh-sloan: Yeah I think you're right we would never match multiple funds here, I think I did it as the call to `get_all_funds()` is above it so we are filtering it. But looking at tit now this would seem like a better option:
`if visible_fund_short_name:
    funds_to_show = get_fund_by_short_name()
else:
    funds_to_show = get_all_funds()`

Then line 43 onwards. Benefit of a fresh pair of eyes :)

PR #305, opened 2023-04-11
bau fail gracefully on incorrect request

gio-karunakaran:
After releasing the multifund changes in frontend where the URL string now directs the user to the respective landing page for the fund and round, we have had misuse of this where requests have ben made using non-existent fund and rounds by editing the URL string directly. 
This makes a call to the fund store which returns a 404, however the get_data() function in frontend forces a 500 and aborts by default if no data was found when making the call.
Another function called get_data_or_fail_gracefully() has been added which is a direct copy of the above get_data() function but does not force the 500 and instead returns a 404. This means we can use this functionality in the future without compromising existing functionality that might depend on the abort 500 returned by get_data() when no data is returned.

### DIFF
--- a/app/default/data.py
+++ b/app/default/data.py
@@ -56,6 +56,39 @@ def get_data(endpoint: str, params: dict = None):
     return data
 
 
+def get_data_or_fail_gracefully(endpoint: str, params: dict = None):
+    """
+        Queries the api endpoint provided and returns a
+        data response in json format. Does not return a 
+        500 on failure but a 404.
+
+    Args:
+        endpoint (str): an API get data address
+
+    Returns:
+        data (json): data response in json format
+    """
+
+    query_string = ""
+    if params:
+        params = {k: v for k, v in params.items() if v is not None}
+        query_string = urlencode(params)
+        endpoint = endpoint + "?" + query_string
+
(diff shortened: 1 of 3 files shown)


app/default/data.py
```diff
@@ -56,6 +56,39 @@ def get_data(endpoint: str, params: dict = None):
...
+        query_string = urlencode(params)
+        endpoint = endpoint + "?" + query_string
+
+    if Config.USE_LOCAL_DATA:
+        current_app.logger.info(f"Fetching local data from '{endpoint}'.")
+        data = get_local_data(endpoint)
+    else:
+        current_app.logger.info(f"Fetching data from '{endpoint}'.")
+        data = get_remote_data(endpoint)
+    if data is None:
```
Adam-W1, line 84: I wonder if we might be losing some valuable insight here, if the API fails it might be useful to read and record the error returned

Adam-W1 (approved):
nice patch 👍

app/default/data.py
gio-karunakaran: added logging of errors from service being called 👍🏾

Adam-W1 (approved):

PR #302, opened 2023-03-30
Fs 2432 start page

srh-sloan:
https://digital.dclg.gov.uk/jira/browse/FS-2432
https://digital.dclg.gov.uk/jira/browse/FS-2183
https://app.mural.co/t/mhclg2837/m/mhclg2837/1675177182644/578e9019ed4a6da033b5de2183289ca7298e5f12?sender=u94fce4b3d87aef5380039386

### Change description
These changes are to make the fund start page work for any fund, based on URL path. The existing index page (`/`) now redirects to `/cof/r2w3` as COF R2W3 is still live and uses that as it's start page. However all funds that exist in the fund store will now be accessible at `/fund_short_name/round_short_name` eg. `/cof/r2w2` or `/COF/R2W3`.

If you try and access this using just a fund short code (eg. `/cof/`) you will get redirected to the most recent round in that fund.

Added a new template `fund_start_page.html` which parameterises all the text on the start page and these values are now pulled from the fund store.

- [x] Unit tests and other appropriate tests added or updated
- [x] README and other documentation has been updated / added (if needed)
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
Run locally with the docker runner. Navigate to the following URLs
- http://localhost:3008/ - should render as currently
- http://localhost:3008/cof/r2w3 - should render as above
- http://localhost:3008/cof/r2w2 - should display Round 2 Window 2 as closed
- http://localhost:3008/cof - redirects to r2w3 page above (As that's the most recently open window)
- http://localhost:3008/made_up_fund - 404

### Screenshots of UI changes (if applicable)
No visible changes to users

PR #309, opened 2023-04-17
Privacy Notice link in footer is dynamic and has fallback

AbuNavsa:
###FS-2429 Make privacy notice url dynamic in footer

### Change description
Previously, frontend footer links for Privacy notice was hardcoded url, this now has changed to retrieve the url from the fund data endpoint and redirects when visiting /privacy url. It also has a fallback of a hardcoded privacy notice url when failing to retrieve from endpoint.

### DIFF
--- a/app/default/content_routes.py
+++ b/app/default/content_routes.py
@@ -71,10 +71,10 @@ def privacy():
     if privacy_notice_url:
         current_app.logger.warning("Privacy notice configured for fund")
         return redirect(privacy_notice_url)
-
-    current_app.logger.warning(
-        "No privacy notice configured for fund. Redirecting..."
-    )
-    return redirect(
-        "https://www.gov.uk/government/publications/community-ownership-fund-privacy-notice/community-ownership-fund-privacy-notice"
-    )
+    else:
+        current_app.logger.warning(
+            "No privacy notice configured for fund. Redirecting..."
+        )
+        return redirect(
+            "https://www.gov.uk/government/publications/community-ownership-fund-privacy-notice/community-ownership-fund-privacy-notice"
+        )
--- a/app/default/content_routes.py
+++ b/app/default/content_routes.py
@@ -29,7 +29,7 @@ def all_questions(fund_short_name, round_short_name):
     if not round:
         return abort(404)
     return render_template(
-        "cof_r2_all_questions.html", round_title=round["title"]
+        "cof_r2_all_questions.html", round_title=round.title
     )
 
 
--- a/app/default/content_routes.py
+++ b/app/default/content_routes.py
@@ -66,11 +66,11 @@ def privacy():
         Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID
     )
 
-    privacy_notice_url = None
-    try:
-        privacy_notice_url = round_data.privacy_notice
-        return redirect(privacy_notice_url)
+    privacy_notice_url = round_data.privacy_notice if hasattr(round_data, 'privacy_notice') else None
 
-    except AttributeError:
-        print("Error finding privacy notice for fund, redirecting")
+    if privacy_notice_url:
+        current_app.logger.warning("Privacy notice configured for fund")
+        return redirect(privacy_notice_url)
+    else:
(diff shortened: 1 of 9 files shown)


app/default/content_routes.py
```diff
@@ -54,3 +57,27 @@ def contact_us():
...
+    current_app.logger.info("Privacy_notice page loaded.")
+    round_data = get_round_data_fail_gracefully(
+        Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID
+    )
+
+    privacy_notice_url = (
+        round_data.privacy_notice
+        if hasattr(round_data, "privacy_notice")
+        else None
+    )
```
tferns, line 73: This may be cleaner:

```suggestion
    privacy_notice_url = getattr(round_data, "privacy_notice", None)
```

app/default/content_routes.py
```diff
@@ -54,3 +57,27 @@ def contact_us():
...
+
+    if privacy_notice_url:
+        return redirect(privacy_notice_url)
+    else:
+        current_app.logger.warning(
+            "No privacy notice configured for fund. Redirecting..."
+        )
+        return redirect(
+            "https://www.gov.uk/government/publications/community-ownership-fund-privacy-notice/community-ownership-fund-privacy-notice"
+        )
```
tferns, line 83: Could probably leave the else since we're using a return early pattern, would leave out a layer of indentation (personal preference, not a big deal):
```suggestion
    
    current_app.logger.warning(
        "No privacy notice configured for fund. Redirecting..."
    )
    return redirect(
        "https://www.gov.uk/government/publications/community-ownership-fund-privacy-notice/community-ownership-fund-privacy-notice"
    )
```

tferns (approved):
👍 some small comments but looks good. 

You have failing unit tests however, so might want to fix them!

app/default/data.py
```diff
@@ -229,7 +229,21 @@ def get_round_data_fail_gracefully(fund_id, round_id):
...
+            fund_id="",
+            opens="",
+            title="",
+            short_name="",
+            prospectus="",
+            privacy_notice="",
+            instructions="",
+            contact_details={},
+            support_availability={},
+        )
```
tferns, line 246: Maybe we could make this a class method of round, so it could be easily re-used elsewhere if need be?

i.e in https://github.com/communitiesuk/funding-service-design-frontend/blob/3e02a7f54fd546d934ba858ab3de4454018e1826/app/models/round.py#L58
```python
@classmethod
@property
def empty_round() -> Round:
    return Round(
            id="",
            assessment_criteria_weighting=[],
            assessment_deadline="",
            deadline="",
            fund_id="",
            opens="",
            title="",
            short_name="",
            prospectus="",
            privacy_notice="",
            instructions="",
            contact_details={},
            support_availability={},
        )
```

```suggestion
        return Round.empty_round
```

app/default/content_routes.py
```diff
@@ -54,3 +57,24 @@ def contact_us():
...
     current_app.logger.info("Cookie policy page loaded.")
     return render_template("cookie_policy.html")
+
+
+@content_bp.route("/privacy", methods=["GET"])
+def privacy():
+    current_app.logger.info("Privacy_notice page loaded.")
+    round_data = get_round_data_fail_gracefully(
+        Config.DEFAULT_FUND_ID, Config.DEFAULT_ROUND_ID
+    )
```
tferns, line 67: I think we're moving to support multi-funds, it would be good to parametrize these instead of assuming it's COF from config. 

I.e see: https://github.com/communitiesuk/funding-service-design-frontend/pull/302

https://github.com/communitiesuk/funding-service-design-frontend/blob/447cba73d2e3f42e759b3b2f74546db684f4ac42/app/default/routes.py#L26-L27

tferns (approved):
Changes look good!  Some suggestions however 👍

app/default/content_routes.py
tferns: CC: @gio-karunakaran @srh-sloan I believe you paired on this, so may be interested in this!

rayhan15243 (approved):

PR #326, opened 2023-05-25
 FS-2819: Redirect to dashboard when clicked back from submit

RamuniN:
https://digital.dclg.gov.uk/jira/browse/FS-2819

### Description
Added cache control, disable going back to tasklist from submit page instead redirect to dashboard

### DIFF
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -211,6 +211,16 @@ def create_app() -> Flask:
             privacy_url=url_for("content_routes.privacy"),
         )
 
+    @flask_app.after_request
+    def after_request(response):
+        if "Cache-Control" not in response.headers:
+            response.headers[
+                "Cache-Control"
+            ] = "no-cache, no-store, must-revalidate"
+            response.headers["Pragma"] = "no-cache"
+            response.headers["Expires"] = "0"
+        return response
+
     @flask_app.before_request
     def filter_favicon_requests():
         if request.path == "/favicon.ico":
--- a/app/default/application_routes.py
+++ b/app/default/application_routes.py
@@ -4,7 +4,6 @@ from http.client import METHOD_NOT_ALLOWED
 import requests
 from app.constants import ApplicationStatus
 from app.default.data import determine_round_status
-from app.default.data import get_account
 from app.default.data import get_application_data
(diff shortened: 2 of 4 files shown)


gio-karunakaran (approved):
LGTM

app/create_app.py
```diff
@@ -211,6 +211,16 @@ def inject_content_urls():
...
 
+    @flask_app.after_request
+    def after_request(response):
+        if "Cache-Control" not in response.headers:
+            response.headers[
+                "Cache-Control"
+            ] = "no-cache, no-store, must-revalidate"
+            response.headers["Pragma"] = "no-cache"
+            response.headers["Expires"] = "0"
+        return response
```
tferns, line 222: IIRC didn't we recently remove this?  Or was that for assessment 🤔

app/default/application_routes.py
```diff
@@ -131,15 +130,6 @@ def tasklist(application_id):
...
     application = get_application_data(application_id, as_dict=True)
-    account = get_account(account_id=application.account_id)
-    if application.status == ApplicationStatus.SUBMITTED.name:
-        with force_locale(application.language):
-            return render_template(
-                "application_submitted.html",
-                application_id=application.id,
-                application_reference=application.reference,
-                application_email=account.email,
-            )
```
tferns, line 142: If I'm reading this right, we don't want to show the user the successful submission page anymore?

And instead just show them their dashboard?

tferns: Oh wait.. this is on the tasklist route 🤔 

Okay I understand the fix now!  Strange.

tferns (approved):
LGTM 🚢

app/create_app.py
RamuniN: cache-control was not implemented in front-end before

app/default/application_routes.py
RamuniN: Here we are preventing the user going back to tasklist page when clicked on back button from submitted page, instead redirect to dashboard

app/create_app.py
tferns: Ah so it was assessment! thanks

PR #327, opened 2023-05-26
Fund_short_name, round_short_name passed to args correctly

AbuNavsa:
Fund and round short name passed to task list page allowing these pages to configure test according to the arguments in the url


tferns (approved):
LGTM 👍

tferns (approved):

PR #331, opened 2023-05-30
BAU: Removed visual index print

Adam-W1:
Remove index printing in favour of configured section names in the fund store
Supports new project-name-ns section, see: https://github.com/communitiesuk/funding-service-design-fund-store/pull/85


tferns (approved):
Looks good but not why sure actions has a red cross, might be some failed tests? 

Looks like it's running so we'll see afterwards

PR #332, opened 2023-05-30
BAU: Log stack trace on errors

tferns:
- Add stacktrace to error logs so we can track down where each error happened

image


adamdavies1 (approved):
LGTM

PR #339, opened 2023-05-31
fs-2856-contact-us-header-fix

gio-karunakaran:
Adding login_requested decorater onto contact_us() route to all flask global variable g to be used with the is.authentcated attribute which was previously not being read as the decorater is what introduces this attribute.

### DIFF
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -172,7 +172,9 @@ def create_app() -> Flask:
         if fund:
             service_title = gettext("Apply for") + " " + fund.title
         else:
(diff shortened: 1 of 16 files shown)


ramsharma-prog (approved):

adamdavies1 (approved):
LGTM. let get Rayhan to give the forms a good go

rayhan15243 (approved):

ramsharma-prog (approved):

ramsharma-prog (approved):

PR #357, opened 2023-06-12
use default round contacts when round name is incorrect

RamuniN:
https://digital.dclg.gov.uk/jira/browse/FS-2978

### Description
Use default round contact details in contact us page when provided round_short_name is incorrect

### DIFF
--- a/app/default/content_routes.py
+++ b/app/default/content_routes.py
@@ -1,4 +1,5 @@
 from app.default.data import get_application_data
+from app.default.data import get_default_round_for_fund
 from app.default.data import get_fund_data
 from app.default.data import get_fund_data_by_short_name
 from app.default.data import get_round_data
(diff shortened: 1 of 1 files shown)


Adam-W1 (commented):
looks good @RamuniN 👍 is this covered in a unit test?

rayhan15243:
@RamuniN Add some unit tests then should be good to go.

Can you confirm that this PR covers all error pages e.g. 404s and 500? I think that's the only two error pages that we have that are customisable?

RamuniN:
> @RamuniN Add some unit tests then should be good to go.
> 
> Can you confirm that this PR covers all error pages e.g. 404s and 500? I think that's the only two error pages that we have that are customisable?

currently we only show contact details in 404s & contact us page only. 500 page doesn't display any contact info. I have implemented the fix in 404s & contact us page to work with correct fund & incorrect round_id

rayhan15243 (approved):

PR #366, opened 2023-08-03
Fs 2877 cached fund round

srh-sloan:
https://dluhcdigital.atlassian.net/browse/FS-2877

### Change description
- Implemented LRU caching for `get_fund_data` & `get_round_data` functions. This reduces the API calls to fund store & improves the performance
- Caching timeout is set to 300 secs (=5 mins)
- Unit tests added for the LRU caching feature.
- Reduce duplicate and redundant code that does the same thing with the same error handling/fallbacks

### How to test
- Put breakpoints in the get_fund and get_round methods of fund_store
- Load the frontend in a browser and check that the above breakpoints don't get hit except on the first retrieval of a session (ie the frontend has cached the data)

### Screenshots of UI changes (if applicable)
n/a


app/default/application_routes.py
```diff
@@ -134,13 +138,15 @@ def tasklist(application_id):
     fund_data = get_fund_data(
         fund_id=application.fund_id,
         language=application.language,
-        as_dict=True,
+        as_dict=False,
```
aaronwilliamsv1, line 141: Why is this now false?

app/default/application_routes.py
```diff
@@ -283,10 +289,14 @@ def submit_application():
...
```
aaronwilliamsv1, line 292: Why is this now false?

app/default/data.py
```diff
@@ -22,6 +24,11 @@
     current_datetime_before_given_iso_string,
 )
 
+
```
aaronwilliamsv1, line 27: So you're looking to get a ttl hash for...5 minutes?

app/default/data.py
```diff
@@ -194,7 +205,11 @@ def get_fund_data_by_short_name(fund_short_name, as_dict=False):
         return Fund.from_dict(fund_response)
 
 
-def get_round_data(fund_id, round_id, language=None, as_dict=False):
+@lru_cache(maxsize=5)
+def get_round_data(
+    fund_id, round_id, language=None, as_dict=False, ttl_hash=None
```
aaronwilliamsv1, line 210: For consistency, these should probably be separated onto their own lines, as [seen below](https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR242)

app/default/data.py
```diff
@@ -168,17 +175,21 @@ def get_applications_for_account(account_id, as_dict=False):
     )
 
 
-def get_fund_data(fund_id, language=None, as_dict=False):
+@lru_cache(maxsize=5)
+def get_fund_data(fund_id, language=None, as_dict=False, ttl_hash=None):
```
aaronwilliamsv1, line 179: For consistency, these should probably be separated onto their own lines, as [seen below](https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR242)

app/default/data.py
```diff
@@ -168,17 +175,21 @@ def get_applications_for_account(account_id, as_dict=False):
...
-        return Fund.from_dict(fund_response)
-    else:
         return fund_response
+    else:
+        return Fund.from_dict(fund_response)
 
 
-def get_fund_data_by_short_name(fund_short_name, as_dict=False):
+@lru_cache(maxsize=5)
+def get_fund_data_by_short_name(fund_short_name, as_dict=False, ttl_hash=None):
```
aaronwilliamsv1, line 191: For consistency, these should probably be separated onto their own lines, as [seen below](https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR242)

app/default/content_routes.py
```diff
@@ -27,15 +27,16 @@ def all_questions(fund_short_name, round_short_name):
...
-    if not round:
-        return abort(404)
 
     if fund_short_name.lower() == "cof":
-        return render_template(
-            f"{fund_short_name.lower()}_{round_short_name.lower()[0:2]}_all_questions.html",
-            round_title=round.title,
-        )
+        if round := get_round(
+            fund_short_name=fund_short_name, round_short_name=round_short_name
```
aaronwilliamsv1, line 33: For consistency, these should probably be separated onto their own lines, as [seen below](https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR242)

app/default/account_routes.py
```diff
@@ -146,11 +150,10 @@ def dashboard():
         # find and display applications with this
         # fund and round else return 404
         template_name = "dashboard_single_fund.html"
-        round_details = get_round_data_by_short_names(
-            fund_short_name,
-            round_short_name,
+        fund_details, round_details = get_fund_and_round(
+            fund_short_name=fund_short_name, round_short_name=round_short_name
```
aaronwilliamsv1, line 154: For consistency, these should probably be separated onto their own lines, as [seen below](https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR242)

app/models/fund.py
```diff
@@ -1,11 +1,5 @@
 import inspect
 from dataclasses import dataclass
-from enum import Enum
```
aaronwilliamsv1, line 3: Are these not needed anymore?

app/default/application_routes.py
RamuniN: Because we want it to return `Fund` object. Function implemented incorrectly previously https://github.com/communitiesuk/funding-service-design-frontend/pull/366/files#diff-25ffeea551d125947892e07d498380fe030e5f7c005ba15e165622de5d567f0eR184

app/default/data.py
RamuniN: yes that's right

app/default/data.py
RamuniN: we don't explicitly separate the lines for small number of argument in a function. --max-line-length=120 is set in hooks, the consistency is handled by the pre-commit hooks

app/models/fund.py
RamuniN: These are removed by Sarah. Basically hardcoded fund-short-names are removed here

config/envs/default.py
```diff
@@ -186,3 +186,6 @@ class DefaultConfig:
     REDIS_INSTANCE_URI = getenv("REDIS_INSTANCE_URI", "redis://localhost:6379")
     TOGGLES_URL = REDIS_INSTANCE_URI + "/0"
     FEATURE_CONFIG = CommonConfig.dev_feature_configuration
+
+    # LRU cache settings
+    LRU_CACHE_TIME = 300  # in seconds
```
aaronwilliamsv1, line 191: Add this setting to uat/production environment to be 3600 - 5 minutes on dev is fine, 1 hour on uat and production makes sense

aaronwilliamsv1 (approved):
Thanks for the changes to the cache time environment - LGTM

PR #384, opened 2023-09-07
fs-3432 added label element

srh-sloan:
### Change description
Minor change to change a `div` to a `label` and fix a critical accessibility bug

### How to test
Launch feedback survey, navigate to page 4 and check that the text for the hours spent field is a label element

Screenshot 2023-09-07 at 11 25 40


RamuniN (approved):
👍

PR #405, opened 2023-10-10
BAU: Don't log errors on 401, just warn

tferns:
- Change this log to a warning, so we don't get spammed in slack anytime someone 401s


RamuniN (approved):
Great One!

PR #412, opened 2023-10-16
FS-3428 making all questions route render correct template for welsh

srh-sloan:
## Change description
Updates the all questions route to deal with the language code being in the filename for all welsh funds after COF R3.

The existing method of having a welsh version of the all questions page is to have translation tags in the template, and call the correct string from babel at runtime.
Now the all questions pages are generated from the forms, we just generate a page for english and a page for welsh, so this method determines which template to render based on language code.

 

- [ ] Unit tests and other appropriate tests added or updated
- [x]  README and other documentation has been updated / added (if needed)
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

## How to test
Visit the all questions page for all exisiting funds/rounds (COF, NSTF, DPIF, CYP) and check it renders in the expected language


app/default/content_routes.py
```diff
@@ -33,9 +34,25 @@ def all_questions(fund_short_name, round_short_name):
...
+        lang = get_lang()
+
+        # Allow for COF R2 and R3 to use the old mechanism for translating all questions into welsh - the template
+        # for these rounds contains translation tags to build the page on the fly.
+        # All future rounds that need welsh all questions will have them generated from the form json so should
+        # be named with the language in the filename
+        all_questions_prefix = (
+            f"{fund_short_name.lower()}_{round_short_name.lower()[0:2]}"
+        )
+        template_name = f"{all_questions_prefix}_all_questions_{lang}.html"
```
tferns, line 46: Hesitant about the `[0:2]`, for COF which has `R2W2` & `R2W3`  - does this logic apply to both windows, and what about future windows?  Would it be better to use the full round_short_name to clear up this confusion?  Would need to re-name the files though.

app/default/content_routes.py
```diff
@@ -33,9 +34,25 @@ def all_questions(fund_short_name, round_short_name):
...
 
     if fund and round:
+        lang = get_lang()
+
+        # Allow for COF R2 and R3 to use the old mechanism for translating all questions into welsh - the template
+        # for these rounds contains translation tags to build the page on the fly.
+        # All future rounds that need welsh all questions will have them generated from the form json so should
+        # be named with the language in the filename
+        all_questions_prefix = (
+            f"{fund_short_name.lower()}_{round_short_name.lower()[0:2]}"
```
tferns, line 44: i.e:

```suggestion
            f"{fund_short_name.lower()}_{round_short_name.lower()}"
```

app/default/content_routes.py
```diff
@@ -33,9 +34,25 @@ def all_questions(fund_short_name, round_short_name):
...
+        )
+        template_name = f"{all_questions_prefix}_all_questions_{lang}.html"
+        if fund.welsh_available and all_questions_prefix.startswith(
+            ("cof_r2", "cof_r3")
```
tferns, line 48: i.e:

```suggestion
            ("cof_r2w2", "cof_r2w3", "cof_r3w2", "cof_r3w1")
```

tferns (approved):
Nice one, I think what you've got works, but I have some concerns for future COF rounds, also - maybe it would be good to have a directory specifically for the all questions html files instead of them just inside `templates`, something like `templates/all_questions/`?

app/default/content_routes.py
srh-sloan: So this is the existing logic and is there because all windows in the same round have the same set of questions (so far anyway)

srh-sloan: So if we renamed we'd also have to duplicate files

tferns: Ah I see, makes sense, guess my worry is we have a new window with different content and it by default uses this - though I think we'd handle that during the development/introduction of the new rounds.

tferns (approved):

PR #496, opened 2023-12-13
Correct language cookie if incorrect

AbuNavsa:
FS-3688 FS-3690 FS-3691 - Language cookie correction on tasklist load

### Change description
Before, if there is a welsh and english application, if the language cookie was set to the wrong language, when entering the application, the cookie would still be the incorrect language and so some translations would show incorrectly and the Feedback survey which is static, would also show incorrectly


SophieOHagan99 (approved):
LGTM

PR #521, opened 2024-01-22
FS-4005 use custom werkzeug in requirements

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-4005

### Description
Use forked `werkzeug` package in the requirements
linked PR https://github.com/communitiesuk/funding-service-design-docker-runner/pull/55

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,19 @@
 FROM python:3.10-bullseye
+ARG USE_DEV_REQUIREMENTS
 
 WORKDIR /app
+
+COPY requirements.txt requirements.txt
 COPY requirements-dev.txt requirements-dev.txt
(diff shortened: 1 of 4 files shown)


srh-sloan (changes requested):
All good but I think you need to change requirements-dev.in as well otherwise it will get overwritten next time someone runs pip-compile

RamuniN:
> All good but I think you need to change requirements-dev.in as well otherwise it will get overwritten next time someone runs pip-compile

Done 👍

srh-sloan (approved):

gidsg:
@srh-sloan @RamuniN FYI, think it would be safe to use requirements-dev in the Dockerfile. As on the deployed environments we use Paketo which doesn't make use of that Dockerfile.

PR #532, opened 2024-02-08
removing unused workflow file and deprecated vscode settings

srh-sloan:
BAU - there's a dependabot PR against a workflow we don't use


robk-dluhc (approved):
🧀

PR #536, opened 2024-02-12
added EOI submitted pages

RamuniN:
https://dluhcdigital.atlassian.net/browse/FS-4029
https://dluhcdigital.atlassian.net/browse/FS-4036
https://dluhcdigital.atlassian.net/browse/FS-4080

### Description
- Added EOI submission page
- Added helper function to evaluate EOI response

### Screenshots of UI changes (if applicable)
image


app/default/application_routes.py
```diff
@@ -346,20 +348,35 @@ def submit_application():
...
+            return render_template(
+                "eoi_submitted.html",
+                eoi_pass=eoi_results["decision"] in [Eoi_Decision.PASS, Eoi_Decision.PASS_WITH_CAVEATS],
+                fund_name=fund_data.name,
+                round_name=round_data.title,
+                fund_short_name=fund_data.short_name,
+                round_short_name=round_data.short_name,
+                round_prospectus=round_data.prospectus,
+                migration_banner_enabled=Config.MIGRATION_BANNER_ENABLED,
+            )
```
tferns, line 363: I'm guessing in the future we'll have some way to mark a fund as EOI, instead of having to check the short name exists inside a tuple/list?

tferns (approved):
Looks good just one comment for future consideration

app/default/application_routes.py
RamuniN: just want to make the code generic if we have EOI for other rounds(e.g `NSTF-EOI`) then in that case we just append it to the tuple.

PR #565, opened 2024-03-19
Fs 4294 shared determine jobs

srh-sloan:
Updated copilot_deploy workflow to use shared workflow for determining jobs to run


Tiny49 (approved):
LGTM

PR #636, opened 2024-10-10
fix notification url issue

nuwan-samarasinghe:
fix notification url issue


sfount (approved):
Good to have the context in the git comment for why this breaks and fixes things but I think I can see this makes the new URL non-default. Happy to get it through!

PR #642, opened 2024-10-21
Simplify error handlers

albertkol:
### Change description

Few things that changed for the 404 error handler:
1. Change create_app to use `utility_processor` for injecting `service_title` - now it only gets executed if invoked, rather than every request.
2. Change `base.py` to default to static title if `is_error` exists and is `True`
3. Simplify the 404 error_handler to prevent potential error cascades

Things that changed for the 401 error handler:
1. Deleted all the logic from the error handler. It was passing data to the 500.html and not doing anything with it.

### How to test
Are tests passing?


app/default/error_routes.py
```diff
@@ -21,26 +21,8 @@
 @account_bp.errorhandler(404)
 def not_found(error):
     current_app.logger.warning(f"Encountered 404 against url {request.path}: {error}")
```
samuelhwilliams, line 23: Can we drop this warning message as well? 🙏

app/create_app.py
```diff
@@ -114,22 +114,26 @@ def inject_global_constants():
...
+                        {request.view_args}, and args: {request.args}"""
+                    )
+            if fund:
+                service_title = gettext("Apply for") + " " + fund.title
+            else:
+                service_title = gettext("Access Funding")
+
+            return service_title
+
+        return dict(service_title=inject_service_name())
```
samuelhwilliams, line 136: This still just gets executed eagerly/immediately, doesn't it?

albertkol: I have to fix this. By invoking the function there, it makes it act exactly as before (on every request). 
Instead, I should just pass the function. 

Had it like that initially, but then changed it without checking again.

samuelhwilliams: > I have to fix this. By invoking the function there, it makes it act exactly as before (on every request). Instead, I should just pass the function.

This reads exactly like chatgpt 😂

app/templates/header.html
```diff
@@ -23,7 +23,7 @@
             </a>
         </div>
         <div class="govuk-header__content">
-            <span class="app-service-name govuk-header__link--service-name">{{ service_title }}</span>
+            <span class="app-service-name govuk-header__link--service-name">{{ "Access Funding" if is_error else service_title }}</span>
```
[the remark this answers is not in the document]
albertkol: Even if that worked, would it not be awfully confusing to troubleshoot or decipher? As `service_title` could come from two different places. Not saying that `is_error` is the prettiest solution either.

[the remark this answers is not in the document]
samuelhwilliams: Potentially confusing, I guess that's fair, yeah.

app/create_app.py
albertkol: I have achieved my true potential. expandingbrain.jpg

app/templates/header.html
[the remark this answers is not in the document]
sfount: 👍 I was thinking this but it is how a lot of the existing app is set up.

Having an individual handler override the value wouldn't give you an indication looking at the templates that it could ever be anything other than `service_title` which refers to the individual funds elsewhere.

I don't feel like there are enough shared requirements between the main app and these extra pages to try and fit them all into one -- I've previously just used different base templates to inherit from (i.e "base/logged in", "base/error", "base/logged out") but that benefits from generic components that can be used in multiple base templates and easily overriden. Looking at things like the header macro it looks there's a lot of delicate logic going on in there and I don't think its worth abstracting it out for this change.

app/default/error_routes.py
albertkol: Deleted!

app/create_app.py
albertkol: Fixed it to invoke the function in the template.

app/templates/header.html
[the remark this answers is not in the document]
albertkol: That's a lot of words just to avoid saying that you agree with my solution.

`service_title` was overwritten anyway by one of the `view` functions. I did an ugly `or` now to echo the view context before getting the `service_title` from the `context_processor`.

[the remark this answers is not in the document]
albertkol: Used `gettext`.

albertkol:
Squeezed in the 401 error handler simplification in this PR.

samuelhwilliams:
shall re-review tomorrow (nudge me if i forget - cheers)

app/create_app.py
```diff
@@ -114,22 +114,25 @@ def inject_global_constants():
...
+        def _get_service_title():
+            fund = None
+            if request.view_args or request.args or request.form:
+                try:
+                    fund = find_fund_in_request()
+                except Exception as e:  # noqa
+                    current_app.logger.warning(
+                        f"""Exception: {e}, occured when trying to
+                        reach url: {request.url}, with view_args:
+                        {request.view_args}, and args: {request.args}"""
```
samuelhwilliams, line 127: The triple-quote string here is weird (but I recognise that's a carryover 👀)

albertkol: Changed it to use %s

app/templates/header.html
```diff
@@ -23,7 +23,7 @@
             </a>
         </div>
         <div class="govuk-header__content">
-            <span class="app-service-name govuk-header__link--service-name">{{ service_title }}</span>
+            <span class="app-service-name govuk-header__link--service-name">{{ gettext("Access Funding") if is_error else service_title or get_service_title() }}</span>
```
gidsg, line 26: What's going on here in this cascade? When is `service_title` defined rather than calling `get_service_title()`? Does it get cached?

albertkol: Not the prettiest situation.

`@application_bp.route("/tasklist/<application_id>", methods=["GET"])` provides a `service_title` to the template.
(no idea why, I just maintained the existing logic)

Before, it used to override the `service_title` from the `context_processor`, as they had the same variable name. And it worked fine, because both were strings.

Now we cannot do that anymore because the `utility_processor` provides a function that we need to invoke in the template.

gidsg (approved):
LGTM 👍 😅

PR #649, opened 2024-11-07
Read maintenance mode variables from parameter store

samuelhwilliams:
This will allow us to turn maintenance mode on/off with a config change and re-deploy, rather than with a code change in a PR. It also means that we can put individual environments into maintenance mode, rather than requiring all envs go into maintenance mode at once.

Would like to pair with whoever reviews this on adding the vars to each environment. Should simply be running these commands for each env:

```
for env in "dev" "test" "uat" "prod"; do
  copilot secret init --name FSD_FRONTEND_MAINTENANCE_MODE --values ${env}=False
  copilot secret init --name FSD_FRONTEND_MAINTENANCE_END_TIME --values ${env}=soon
done
```

https://mhclgdigital.atlassian.net/browse/FSPT-69


srh-sloan (dismissed):

samuelhwilliams:
@srh-sloan apologies, had to refactor a bit - can't have parameter store values that are empty, so tweaked the setting slightly.

srh-sloan (approved):

samuelhwilliams:
Tested in dev and works as expected ✅

srh-sloan (approved):

PR #686, opened 2024-12-06
Switch from black+flake8 to ruff

samuelhwilliams:
### Change description
Apply ruff formatting changes across the codebase, prior to lift+shift.

### DIFF
--- a/app/default/data.py
+++ b/app/default/data.py
@@ -210,7 +210,7 @@ def get_application_display_config(fund_id, round_id, language):
     try:
         return [ApplicationMapping.from_dict(section) for section in application_display_response]
     except Exception as e:
-        raise ValueError(f"Failed to create ApplicationMapping instance: {str(e)}")
+        raise ValueError("Failed to create ApplicationMapping instance") from e
 
 
 @lru_cache(maxsize=5)
--- a/app/default/data.py
+++ b/app/default/data.py
@@ -7,25 +7,22 @@ from typing import List
 from urllib.parse import urlencode
 
 import requests
+from flask import abort, current_app
+from fsd_utils.locale_selector.get_lang import get_lang
+from fsd_utils.simple_utils.date_utils import (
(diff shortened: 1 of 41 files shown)


pyproject.toml
```diff
@@ -37,26 +33,45 @@ max_attribute_length=1000   # ^^^
...
+    "C90",  # mccabe cyclomatic complexity
+    "G",  # flake8-logging-format
+]
+ignore = []
+exclude = [
+    "db/migrations/versions/",
+    "venv*",
+    ".venv*",
+    "__pycache__",
+    "fund_store/config/fund_loader_config/FAB/"
```
MarcUsher, line 57: We prob don't need the `db/migrations/versions/` and `fund_store` exceptions here, unless we're just pre-empting the eventual merge?

app/default/data.py
```diff
@@ -210,7 +210,7 @@ def get_application_display_config(fund_id, round_id, language):
     try:
         return [ApplicationMapping.from_dict(section) for section in application_display_response]
     except Exception as e:
-        raise ValueError(f"Failed to create ApplicationMapping instance: {str(e)}")
+        raise ValueError("Failed to create ApplicationMapping instance") from e
```
MarcUsher, line 213: Do we want to include the error or some of its info in the ValueError being raised?

MarcUsher (approved):
Looks good! Approving but have just left 2 tiny 🤏  comments.

app/default/data.py
samuelhwilliams: The info from the previous error gets chained as context into this new error via the `from e`, so the context should be visible from the stacktrace still.

pyproject.toml
samuelhwilliams: Just a lazy copy paste on my side. I'm gonna continue to be lazy and leave them, since we can rationalise in the combination...

